PR #5, opened 2016-01-25
Added support for file globs in source_directories.

ypid:
`source_directories_glob` can be used to enable glob support (defaults to
disabled). Let me know what needs to be changed (option names or other stuff) for this patch to get accepted.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@
  * Fixed handling of repeated spaces in source_directories which resulted in backup up everything.
  * Added support for --one-file-system for Borg.
  * Support borg create --umask.
+ * Added support for file globs in source_directories.
 
 0.1.7
 
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ Here's an example config file:
 ```INI
 [location]
 # Space-separated list of source directories to backup.
-source_directories: /home /etc
-# source_directories_glob: 1
+# Globs are expanded.
+source_directories: /home /etc /var/log/syslog*
 
 # Path to local or remote backup repository.
 repository: user@backupserver:sourcehostname.attic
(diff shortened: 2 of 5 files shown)


witten:
Great idea.

Can you see any reason not to just glob unconditionally? That way, a user can get the benefits of globbing simply by specifying a path containing a "*", without having to also set a boolean option to enable the feature.

ypid:
Not sure. Maybe users have `*` in there file names. I guess we could enable it by default.

witten:
Yeah, I think it's reasonable to say "*" always means glob, especially if it's documented as such.

ypid:
All right. Can you merge my other PRs so that I can rebase to them and updated this one? (I intend to change this parm to boolean as well).

witten:
All merged.

IMO, you can simply drop the `source_directories_glob` flag entirely (boolean or otherwise) and default to always globbing. One less value to configure that way.

ypid:
Changing it to default has a the side effect removing not existing source_directories from the list. `python -c 'from glob import glob; print(glob("not existing"))'`. Borg would otherwise complain about non existing files. Should I change it to include non existing files and let attic/borg handle it?

witten:
Ah, good point. To deal with the non-existent directory case, what do you think of checking if a particular `glob()` call returns an empty list. And if so, erroring. That might look something like:

```
expanded_sources = [glob(x) for x in sources]
if [] in expanded_sources:
    raise ValueError('oops')
flattened_sources = list(chain.from_iterable(expanded_sources))
```

That way, you'd catch both the case of specifying a literal directory that doesn't exist, and of specifying a glob path that doesn't resolve to anything.

ypid:
I updated the PR just now. I was going with just passing non existing files to the backend and let it deal with that.

witten:
Awesome.. That makes sense as well.

witten:
Note that I altered the test in cf545ae93a26bfb2ed023c2407f09d6601ee761a to mock out the glob() call, so it isn't subject to what's present on the filesystem during the test.

PR #6, opened 2016-01-30
Added .gitignore file.

ypid:


witten:
I'm more than happy to add a .gitignore. The fact that there isn't one is an oversight on my part because I mostly use Mercurial for this project. But the proposed .gitignore includes all sorts of file types that don't apply to this project. What do you think of just ignoring the file paths currently mentioned in the .hgignore? Specifically:

```
syntax: glob
*.egg-info
*.pyc
*.swp
.tox
build
dist
```

It's even arguable that `*.swp` shouldn't be among that list, because that's tool-specific (vim) and has nothing to do with this project itself.

ypid:
Sorry, I did not check for `.hgignore`. You are also right about the `*.swp`. I updated the PR.

PR #7, opened 2016-01-30
Added support for --one-file-system for Borg.

ypid:
Note that I already resolved the merge conflicts with #5 in https://github.com/ypid/atticmatic/tree/merge in case you decide to merge both PR.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ Dan Helfman <witten@torsion.org>: Main developer
 
 Alexander Görtz: Python 3 compatibility
 Henning Schroeder: Copy editing
+Robin `ypid` Schneider: Support additional options of Borg
--- a/atticmatic/backends/borg.py
+++ b/atticmatic/backends/borg.py
@@ -11,7 +11,7 @@ CONFIG_FORMAT = (
         'location',
         (
             option('source_directories'),
-            option('one_file_system', required=False),
+            option('one_file_system', value_type=bool, required=False),
             option('repository'),
         ),
     ),
--- a/atticmatic/backends/borg.py
+++ b/atticmatic/backends/borg.py
@@ -7,7 +7,14 @@ from atticmatic.backends import shared
 
 COMMAND = 'borg'
 CONFIG_FORMAT = (
-    shared.CONFIG_FORMAT[0],  # location
+    Section_format(
+        'location',
+        (
+            option('source_directories'),
+            option('one_file_system', required=False),
+            option('repository'),
+        ),
+    ),
     Section_format(
         'storage',
         (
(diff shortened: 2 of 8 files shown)


AUTHORS
```diff
@@ -2,3 +2,4 @@ Dan Helfman <witten@torsion.org>: Main developer
 
 Alexander Görtz: Python 3 compatibility
 Henning Schroeder: Copy editing
+Robin `ypid` Schneider: Support additional options of Borg
```
witten: Ah, great. I see you found the AUTHORS file in this changeset. :)

sample/config
```diff
@@ -2,6 +2,8 @@
 # Space-separated list of source directories to backup.
 source_directories: /home /etc
 
+# one_file_system: 1
```
witten: Maybe add a brief comment above this line describing it?

atticmatic/backends/borg.py
```diff
@@ -7,7 +7,14 @@
 
 COMMAND = 'borg'
 CONFIG_FORMAT = (
-    shared.CONFIG_FORMAT[0],  # location
+    Section_format(
+        'location',
+        (
+            option('source_directories'),
+            option('one_file_system', required=False),
```
witten: It's probably a good idea here to pass in `value_type=bool`. Note that if you do that, you'll have to alter `type_getter` in `parse_section_options()`, adding something like `bool: parser.getboolean,`

sample/config
witten: This example might be clearer if the value was `true`. See: https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.getboolean

atticmatic/tests/unit/backends/test_shared.py
```diff
@@ -131,6 +131,22 @@ def test_create_archive_with_compression_should_call_attic_with_compression_para
     )
 
 
+def test_create_archive_with_one_file_system_should_call_attic_with_one_file_system_parameters():
```
witten: Yay tests.

witten:
Please add this great new feature to the NEWS file!

atticmatic/tests/unit/backends/test_shared.py
ypid: I am just continuing your great work :+1: :wink:

ypid:
You are right. Thanks for the review. Done.

witten:
Manually merged this in 049f9c8853ef287a547e55ab2b3d02ea6b0c6cd7. My only change was altering the version number to make Python's version normalization happy.

Thanks!

ypid:
Thanks. Just a note. Git supports author and committer separately.

witten:
I'll look into that for future merges. I'm using [HgGit](https://www.mercurial-scm.org/wiki/HgGit) as a bridge from Mercurial to GitHub, so it may or may not support that. I suppose I can always use git directly for such merges and then pull them back to my Mercurial repo.

PR #8, opened 2016-01-31
Use /\s+/ to split source_directories to handle 1+ spaces.

ypid:
This bug is can be quite annoying because when you accidentally used
something like:

```ini
[location]
source_directories: backup_one  backup_two
;                              A (Additional space here)
```

It would call Attic/Borg with ('backup_one', '', 'backup_two') which in
turn backups your whole $PWD.

### DIFF
--- a/atticmatic/tests/unit/backends/test_shared.py
+++ b/atticmatic/tests/unit/backends/test_shared.py
@@ -71,6 +71,21 @@ def test_create_archive_should_call_attic_with_parameters():
     )
 
 
+def test_create_archive_with_two_spaces_in_source_directories():
+    insert_subprocess_mock(CREATE_COMMAND)
+    insert_platform_mock()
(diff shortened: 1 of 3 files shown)


atticmatic/tests/unit/backends/test_shared.py
```diff
@@ -71,6 +71,21 @@ def test_create_archive_should_call_attic_with_parameters():
     )
 
 
+def test_create_archive_with_two_spaces_in_source_directories():
```
witten: Thanks for adding a test!

witten:
Can you add an entry for this fix to the NEWS file when you get a chance? You can create a new "dev" version there, since it's not released yet.

Also, feel free to add yourself to the AUTHORS file if you like.

ypid:
Done.
Cheers

PR #9, opened 2016-01-31
Support borg create --umask.

ypid:


atticmatic/backends/borg.py
```diff
@@ -13,6 +13,7 @@
         (
             option('encryption_passphrase', required=False),
             option('compression', required=False),
+            option('umask', int, required=False),
```
witten: Should this be a string instead of an int, to support octal values properly? The example from the [Borg docs](https://borgbackup.readthedocs.org/en/stable/usage.html#borg-create) is `0077`. Taking an int config value and calling `str()` on it would result in just `77`, for instance. Not sure if this matters from Borg's perspective. I assume you've tried it out?

witten:
Thanks for implementing this! Would you mind adding this feature to the NEWS file for the next release?

atticmatic/backends/borg.py
ypid: I did not test `77` but instead something more common like `0700`. Thanks for the review. I agree that using a string for it is better to avoid such problems. Done.

witten:
Merged manually in 0012e0cdea8d4b6b57c50721ad39e946d95559b5. No changes other than resolving conflicts.

Thanks again!

PR #11, opened 2016-04-06
Fix broken link to Borg quickstart

jangondol:


witten:
Thanks! Note that there were a few other links broken similarly in docs and sample config. I fixed those too.

jangondol:
Good catch, thank you for fixing those too!

PR #13, opened 2016-10-26
Support for multiple repositories.

bfabio:


witten:
Thanks for submitting this! Could you talk about the use case that you intend to use it for? For instance, is it that you want extra redundancy, so you want to back up to several different remote locations?

bfabio:
My use case is exactly the one you mentioned.

I need an offsite backup (well, everyone does!) with the same settings as the main one. I prefer not to copy the borg repository over because:

- Using rsync or another program that supports delta transfer (essential) seems like a poor man's implementation of something a backup tool should support out of the box

- rsync uses quite a lot of memory when copying large amount of data, in my experience

- It's yet another tool to test, deploy, maintain and monitor

- No borg check on the offsite repository

- (most importantly) Issues on the main repository will be copied to the offsite one.

The "repository" configuration directive should really be "repositories", but I took the lazy and compatible approach here. If you like the feature, I can cook up a patch which addresses that.

johanfleury:
You could use multiple config files and run borgmatic with `-c`.

If you use cron, you just need a line for each repositories (that's easily deployable) and if you use systemd you can create a multiple instance service (`borgmatic@.service`).

bfabio:
@johanfleury I'm doing exactly in production right now, but that's suboptimal because most of the configuration file is duplicated and when you have to change a directive, you have to change it in two places.

With cron or systemd you should avoid splitting the runs in favor of starting the second backup as soon as the first finishes (sidenote, this results in comically long cron entries), otherwise you'll have to guess how much time a backup takes.

For multi hour runs this is relevant, because you want those to finish during night hours as soon as possible, before your users wake up.

niol:
(Too bad I find out about borgmatic after writing https://github.com/niol/sya)
Maybe what's missing is some kind of configuration imports or inheritance to avoid cut-n-paste.

bfabio:
@witten What's your take on this?

witten:
Sorry for the lengthy delay in getting back to you. I've been.. busy.

I think the general approach / feature makes sense to me now that you've explained the use case. I agree about "repository".. Ideally it would be "repositories", but that would break compatibility with existing configuration files. So probably the thing to do is to support both, and possibly deprecate the singular "repository".

I'm fine with either you taking another pass at this, or I can run with it from here if you prefer. Note that there will need to be corresponding README and test updates in any case.

witten:
Since this PR, I've switched the config file format to YAML. That will be in the next release. Since the config file format is changing anyway, it's an the opportune time to rename "repository" to "repositories". I'll have a look at incorporating the spirit of this PR into master to accomplish that. Thanks for submitting this!

witten:
The next release is available now: borgmatic 1.1.0. It includes the multiple repositories support. Thanks again!

bfabio:
@witten Thank *you*, you did all the work.

PR #14, opened 2017-05-13
Allow spaces in source directories

poirier:
Use a backslash to escape a space to treat it as part of
the directory name and not a separator.

### DIFF
--- a/borgmatic/borg.py
+++ b/borgmatic/borg.py
@@ -13,6 +13,8 @@ from borgmatic.verbosity import VERBOSITY_SOME, VERBOSITY_LOTS
 
 
 COMMAND = 'borg'
+ESCAPED_SPACE = r'\ '
+TEMP_REPLACEMENT_FOR_ESCAPED_SPACE = 'SLASH_ESCAPED_SPACE'
 
 
 def initialize(storage_config, command=COMMAND):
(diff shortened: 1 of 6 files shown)


witten:
Thanks for making this change. Seems like a reasonable thing to do.. I could also see supporting "/quoted paths" at some point.

borgmatic/borg.py
```diff
@@ -28,10 +30,13 @@ def create_archive(
...
-    list of source directories, a local or remote repository path, and a command to run, create an
-    attic archive.
+    list of source directories (spaces escaped with a backslash not treated as separators), a local
+    or remote repository path, and a command to run, create an attic archive.
     '''
-    sources = re.split('\s+', source_directories)
+    # To support escaping spaces in directory name, temporarily replace them
+    temp_source_directories = source_directories.replace(ESCAPED_SPACE, TEMP_REPLACEMENT_FOR_ESCAPED_SPACE)
+    sources = re.split('\s+', temp_source_directories)
+    sources = [s.replace(TEMP_REPLACEMENT_FOR_ESCAPED_SPACE, ' ') for s in sources]
```
witten: You could accomplish the same thing perhaps a little more succinctly with a negative look-behind regular expression. Here's an example:

```
sources = [source.replace('\\ ', ' ') for source in re.split(r'(?<!\\) ', source_directories)]
```

That `"r'(?<!\\) '` matches any non-backslash character followed by a space, but without capturing that preceding character.

More here: https://docs.python.org/3/library/re.html

witten:
Since this PR, I've switched the borgmatic config file format to YAML. (Will be in the next release.) So you'll no longer have to deal with space-separated source directories in config. Thanks for the PR anyway!

witten:
The next release is available now: borgmatic 1.1.0. And I've confirmed that spaces work in source directory names with that version. Thanks!

PR #15, opened 2017-09-02
Added archive-prefix to config.yaml

mlazze:
This PR adds `archive-prefix` to [storage], which is prefixed to the archive name.
Also, the way the current archive name is created ({archive_prefix}{hostname}-{timestamp}) is added to the generated conf comments for easier access.

Storage and prune prefixes are independent and archive-prefix is set by default to ''. This means that it's backwards compatible and no config changes are required for those who like current setup.

Usecase: mostly having the opportunity to use multiple config files on the same host and being able to distinguish the archives

See: https://tree.taiga.io/project/witten-borgmatic/issue/28

---

The alternative is to use a single prefix for both storage and retention.
We would need the archive name to be changed from {hostname}-{timestamp} to {prefix}-{timestamp} with prefix defaulting to {hostname}. So, either:
 1. no breaking changes: [retention].prefix is the source of truth 
 2. breaking changes: [storage].prefix is the source of truth and [retention].prefix is deprecated

I can create another PR if this alternative is preferred.

### DIFF
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -82,11 +82,13 @@ def create_archive(
         VERBOSITY_SOME: ('--info', '--stats',),
         VERBOSITY_LOTS: ('--debug', '--list', '--stats'),
     }.get(verbosity, ())
+    archive_prefix = storage_config.get('archive-prefix', '')
 
     full_command = (
         'borg', 'create',
-        '{repository}::{hostname}-{timestamp}'.format(
+        '{repository}::{archive_prefix}{hostname}-{timestamp}'.format(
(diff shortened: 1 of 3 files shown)


squires:
Dan, this is why I made the format string configurable. People will want to name things their way. The label is more useful to me as host-label-timestamp than as a prefix.

witten:
Fair point. I didn't realize that Borg supported the format-string-style placeholders until I saw https://github.com/witten/borgmatic/pull/16

Interested in your thoughts on that. That might give the flexibility you're suggesting while letting Borg do all the parsing work.

PR #16, opened 2017-09-03
Added storage.archive_name to config

mlazze:
Related to #15 , https://tree.taiga.io/project/witten-borgmatic/issue/28

"Repository (or Archive) URLs, –prefix and –remote-path values support these placeholders" (https://borgbackup.readthedocs.io/en/stable/usage.html#borg-help-placeholders).

So, an even better way than adding a prefix or handling hostname and timestamp inside borgmatic is to use borg's placeholders which accomplishes the same thing as the old 
```
'{hostname}-{timestamp}'.format(...) 
```

This allows the usecase for both a prefix, label, removing timestamp and all of that, simply by changing storage.archive_name to something else, be it `MYPREFIX-{hostname}-{now}`, `{hostname}-MY_LABEL-{now}-{borgversion}` or whatever.

The default archive_name is now 
```
{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}
```
which is the same as the previously used 
```
'{hostname}-{timestamp}'.format(hostname=platform.node(), timestamp=datetime.now().isoformat())
```
but everything depends on borg's implementation of placeholders.

---

I deleted test_create platform and date mocks as hostname and timestamp are not generated in borgmatic, but the placeholders get passed to borg, so this made the mocks useless.

----

I'm closing #15 as this achieves the same goal in a better way.

### DIFF
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -82,15 +82,17 @@ def create_archive(
         VERBOSITY_SOME: ('--info', '--stats',),
         VERBOSITY_LOTS: ('--debug', '--list', '--stats'),
     }.get(verbosity, ())
+    default_archive_name = '{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}'
+    archive_name = storage_config.get('archive_name', default_archive_name)
 
     full_command = (
         'borg', 'create',
-        '{repository}::{hostname}-{timestamp}'.format(
+        '{repository}::{archive_name}'.format(
             repository=repository,
-            hostname=platform.node(),
-            timestamp=datetime.now().isoformat(),
+            archive_name=archive_name,
         ),
     ) + sources + exclude_flags + compression_flags + one_file_system_flags + \
         remote_path_flags + umask_flags + verbosity_flags
 
     subprocess.check_call(full_command)
+
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -1,8 +1,6 @@
-from datetime import datetime
 import glob
 import itertools
 import os
-import platform
 import subprocess
 import tempfile
 
@@ -82,17 +80,16 @@ def create_archive(
         VERBOSITY_SOME: ('--info', '--stats',),
         VERBOSITY_LOTS: ('--debug', '--list', '--stats'),
     }.get(verbosity, ())
-    default_archive_name = '{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}'
-    archive_name = storage_config.get('archive_name', default_archive_name)
+    default_archive_name_format = '{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}'
+    archive_name_format = storage_config.get('archive_name_format', default_archive_name_format)
 
     full_command = (
         'borg', 'create',
-        '{repository}::{archive_name}'.format(
+        '{repository}::{archive_name_format}'.format(
             repository=repository,
-            archive_name=archive_name,
+            archive_name_format=archive_name_format,
         ),
     ) + sources + exclude_flags + compression_flags + one_file_system_flags + \
         remote_path_flags + umask_flags + verbosity_flags
 
     subprocess.check_call(full_command)
-
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -91,10 +91,10 @@ map:
             archive_name:
                 type: scalar
                 desc: |
-                    Name of the archive. Borg placeholders can be used, see
+                    Name of the archive. Borg placeholders can be used. See
                     https://borgbackup.readthedocs.io/en/stable/usage.html#borg-help-placeholders
                     Default is "{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}"
-                example: "Documents-{hostname}-{now}"
+                example: "{hostname}-documents-{now}"
     retention:
         desc: |
             Retention policy for how many backups to keep in each category. See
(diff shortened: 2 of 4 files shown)


borgmatic/tests/unit/borg/test_create.py
```diff
@@ -128,15 +118,13 @@ def test_make_exclude_flags_is_empty_when_config_has_no_excludes():
     assert exclude_flags == ()
 
 
-CREATE_COMMAND = ('borg', 'create', 'repo::host-now', 'foo', 'bar')
+CREATE_COMMAND = ('borg', 'create', 'repo::{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}', 'foo', 'bar')
```
witten: To avoid duplication of the default archive name, you could pull ` default_archive_name` out into a constant, and then reference/interpolate it here. I generally try not to refer to code-under-test in tests themselves, but I make an exception for constants. Same thing for other uses of that archive name format below.

I don't feel super strongly about this particular instance of duplication though.

borgmatic/borg/create.py
```diff
@@ -82,15 +82,17 @@ def create_archive(
         VERBOSITY_SOME: ('--info', '--stats',),
         VERBOSITY_LOTS: ('--debug', '--list', '--stats'),
     }.get(verbosity, ())
+    default_archive_name = '{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}'
```
witten: Nice.. I had no idea Borg supported this placeholder thing.

borgmatic/borg/create.py
```diff
@@ -82,15 +82,17 @@ def create_archive(
...
             repository=repository,
-            hostname=platform.node(),
-            timestamp=datetime.now().isoformat(),
+            archive_name=archive_name,
         ),
     ) + sources + exclude_flags + compression_flags + one_file_system_flags + \
         remote_path_flags + umask_flags + verbosity_flags
 
     subprocess.check_call(full_command)
+
```
witten: Nit: Remove empty line at end of file.

borgmatic/borg/create.py
```diff
@@ -82,15 +82,17 @@ def create_archive(
...
+    default_archive_name = '{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}'
+    archive_name = storage_config.get('archive_name', default_archive_name)
 
     full_command = (
         'borg', 'create',
-        '{repository}::{hostname}-{timestamp}'.format(
+        '{repository}::{archive_name}'.format(
             repository=repository,
-            hostname=platform.node(),
-            timestamp=datetime.now().isoformat(),
```
witten: `platform` and `datetime` are now unused imports, above.

borgmatic/config/schema.yaml
```diff
@@ -88,6 +88,13 @@ map:
                 type: scalar
                 desc: Umask to be used for borg create.
                 example: 0077
+            archive_name:
```
witten: I might suggest calling this `archive_name_format` to make clear that it's the pattern for archive names, rather than the literal archive name itself. (This naming idea is from an earlier patch from @squires.)

borgmatic/config/schema.yaml
```diff
@@ -88,6 +88,13 @@ map:
...
+                type: scalar
+                desc: |
+                    Name of the archive. Borg placeholders can be used, see
```
witten: Nit: ", see" -> ". See"

borgmatic/config/schema.yaml
```diff
@@ -119,7 +126,10 @@ map:
                 example: 1
             prefix:
                 type: scalar
-                desc: When pruning, only consider archive names starting with this prefix.
+                desc: |
+                    When pruning, only consider archive names starting with this prefix.
+                    Borg placeholders can be used, see
```
witten: Nit: ", see" -> ". See"

borgmatic/tests/unit/borg/test_create.py
```diff
@@ -344,3 +312,41 @@ def test_create_archive_with_glob_should_call_borg_with_expanded_directories():
         },
         storage_config={},
     )
+
+
+def test_create_archive_with_archive_name():
```
witten: Thanks for adding automated tests!

However I don't think this particular test adds a whole lot versus the test immediately below. I suppose it would catch a regression in which we treated literal archive names differently from placeholder archive names, but that's it.

If you do want to keep this test, I would recommend making its name clearly reflect what exactly it's testing for. Example: `test_create_archive_with_archive_name_passes_through_to_borg()`.

borgmatic/config/schema.yaml
```diff
@@ -88,6 +88,13 @@ map:
                 type: scalar
                 desc: Umask to be used for borg create.
                 example: 0077
+            archive_name:
+                type: scalar
+                desc: |
+                    Name of the archive. Borg placeholders can be used, see
+                    https://borgbackup.readthedocs.io/en/stable/usage.html#borg-help-placeholders
+                    Default is "{hostname}-{now:%Y-%m-%dT%H:%M:%S.%f}"
+                example: "Documents-{hostname}-{now}"
```
witten: When I think of a label like "Documents", I tend to think of it as logically coming *after* hostname.. My rationale is that the archive name format is sort of a series of namespaces from least specific to most specific. E.g., first hostname, then the label for the thing being backed up on that particular host, and finally the timestamp of when it was backed up. So I would suggest an example like "{hostname}-documents-{now}" or similar.

One implication of that change is that when matching on a prefix (for pruning), the prefix would have to contain either "{hostname}" or "{hostname}-documents". That might be totally okay.

However, I do not feel super strongly about this point.

witten:
Thanks all. I just released this in borgmatic 1.1.7.

squires:
I did not realize borg had placeholders either. This way is a lot easier.

PR #17, opened 2017-10-01
Support for user hooks

Churro:
This PR adds support for user-defined shell scripts or single commands to be executed before and after a backup, or if an error has occurred. The intended use case is just as described in two existing issues [1, 2].  

Using the "before_backup", "after_backup", and "on_error" sections in the config file, it is possible to list one or multiple commands to execute. For example, this can be used to send emails after successful / failed backups, check disk space, execute mysqldump prior to creating a backup, etc.

For a better overview I've split the implementation into 3 commits:

1. af38fdea01a145a6c3437df0193823bb04d20e21 enables the value of the "example" fields in schema.yaml to be false by default, e.g. when used in conjunction with _type: bool_. 
This comes in handy to demonstrate the use of hooks in the generated default config. As the user has to customize the commands to execute anyway, it seems advisable to disable the hooks by default.

2. 55fd81332455270a94392adda3502fafd0b8b655 is the actual implementation for hooks. For now, I've moved _exec_cmd()_ into the separate file hook.py. As it's only 4 lines it might as well fit into borgmatic.py but I guess that's a matter of taste :-)
 
3. 7685613d280817c5efd954c715d12f339fa2753f removes "example" fields recursively from _schema_. It might be a bit overkill just for user hooks but I think it's useful anyway, considering that future config versions could add further "map" sections regardless of their position in the hierarchy.

It may not be missed that shell support in borgmatic also has some security implications:

- In the worst case scenario, a config file specifies a shell script that is also writable by attackers. For example, lets assume the specified script resides in the document root of a vHost user and the provided web application is not exactly secure. After successful exploitation an attacker might place his own commands into the script and wait for the next cron run of borgmatic. In that case, an attacker's payload would be executed with the same permissions as borgmatic. If borgmatic is called with higher-level permissions, the consequences could be fatal.

- In another scenario, imagine config files having insecure file permissions. An attacker couldn't only read the encryption passphrase but might also be able to insert arbitrary shell commands. The implications would be the same as above.

In summary, I think user hooks would be a very useful extension to borgmatic **if used with care**.  Anyone enabling them should understand the security implications. 

[1] https://tree.taiga.io/project/witten-borgmatic/issue/16
[2] https://tree.taiga.io/project/witten-borgmatic/issue/38

### DIFF
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -5,8 +5,8 @@ from subprocess import CalledProcessError
 import sys
 
 from borgmatic.borg import check, create, prune
-from borgmatic.config import collect, convert, validate
 from borgmatic.commands import hook
+from borgmatic.config import collect, convert, validate
 
 
 LEGACY_CONFIG_PATH = '/etc/borgmatic/config'
@@ -75,7 +75,6 @@ def parse_arguments(*arguments):
 
 
 def main():  # pragma: no cover
-    hooks = {}
     try:
         args = parse_arguments(*sys.argv[1:])
         config_filenames = tuple(collect.collect_config_filenames(args.config_paths))
@@ -92,24 +91,27 @@ def main():  # pragma: no cover
             )
             remote_path = location.get('remote_path')
 
-            create.initialize(storage)
-            hook.exec_cmd(hooks.get('before_backup'))
+            try:
+                create.initialize(storage)
+                hook.execute_hook(hooks.get('before_backup'))
 
-            for repository in location['repositories']:
-                if args.prune:
-                    prune.prune_archives(args.verbosity, repository, retention, remote_path=remote_path)
-                if args.create:
-                    create.create_archive(
-                        args.verbosity,
-                        repository,
-                        location,
-                        storage,
-                    )
-                if args.check:
-                    check.check_archives(args.verbosity, repository, consistency, remote_path=remote_path)
+                for repository in location['repositories']:
+                    if args.prune:
+                        prune.prune_archives(args.verbosity, repository, retention, remote_path=remote_path)
+                    if args.create:
+                        create.create_archive(
+                            args.verbosity,
+                            repository,
+                            location,
+                            storage,
+                        )
+                    if args.check:
+                        check.check_archives(args.verbosity, repository, consistency, remote_path=remote_path)
 
-            hook.exec_cmd(hooks.get('after_backup'))
+                hook.execute_hook(hooks.get('after_backup'))
+            except (OSError, CalledProcessError):
+                hook.execute_hook(hooks.get('on_error'))
+                raise
     except (ValueError, OSError, CalledProcessError) as error:
-        hook.exec_cmd(hooks.get('on_error'))
         print(error, file=sys.stderr)
         sys.exit(1)
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -6,6 +6,7 @@ import sys
 
 from borgmatic.borg import check, create, prune
 from borgmatic.config import collect, convert, validate
+from borgmatic.commands import hook
 
 
 LEGACY_CONFIG_PATH = '/etc/borgmatic/config'
@@ -74,6 +75,7 @@ def parse_arguments(*arguments):
 
 
 def main():  # pragma: no cover
+    hooks = {}
     try:
         args = parse_arguments(*sys.argv[1:])
         config_filenames = tuple(collect.collect_config_filenames(args.config_paths))
@@ -84,13 +86,14 @@ def main():  # pragma: no cover
 
         for config_filename in config_filenames:
             config = validate.parse_configuration(config_filename, validate.schema_filename())
-            (location, storage, retention, consistency) = (
+            (location, storage, retention, consistency, hooks) = (
                 config.get(section_name, {})
-                for section_name in ('location', 'storage', 'retention', 'consistency')
+                for section_name in ('location', 'storage', 'retention', 'consistency', 'hooks')
             )
             remote_path = location.get('remote_path')
 
             create.initialize(storage)
+            hook.exec_cmd(hooks.get('before_backup'))
 
             for repository in location['repositories']:
                 if args.prune:
@@ -104,6 +107,9 @@ def main():  # pragma: no cover
                     )
                 if args.check:
                     check.check_archives(args.verbosity, repository, consistency, remote_path=remote_path)
+
+            hook.exec_cmd(hooks.get('after_backup'))
     except (ValueError, OSError, CalledProcessError) as error:
+        hook.exec_cmd(hooks.get('on_error'))
         print(error, file=sys.stderr)
         sys.exit(1)
--- a/borgmatic/commands/hook.py
+++ b/borgmatic/commands/hook.py
@@ -1,7 +1,7 @@
 import subprocess
 
 
-def exec_cmd(config):
-    if config and config.get('enable_hook', None) is True:
-        for cmd in config.get('exec_hook'):
+def execute_hook(commands):
+    if commands:
+        for cmd in commands:
             subprocess.check_call(cmd, shell=True)
--- a/borgmatic/commands/hook.py
+++ b/borgmatic/commands/hook.py
@@ -1,0 +1,7 @@
+import subprocess
+
+
+def exec_cmd(config):
+    if config and config.get('enable_hook', None) is True:
+        for cmd in config.get('exec_hook'):
+            subprocess.check_call(cmd, shell=True)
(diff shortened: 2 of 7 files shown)


witten:
Thanks for doing this! I'll have a look through all of it when I get a chance. But just based on your description, sounds like a very useful addition.

borgmatic/config/generate.py
```diff
@@ -24,7 +24,7 @@ def _schema_to_sample_configuration(schema, level=0):
     for each section based on the schema "desc" description.
     '''
     example = schema.get('example')
-    if example:
+    if example is not None:
```
witten: This change makes sense to me, but it would be nice to have some test coverage it.

borgmatic/config/schema.yaml
```diff
@@ -157,3 +157,52 @@ map:
                 desc: Restrict the number of checked archives to the last n. Applies only to the
                       "archives" check.
                 example: 3
+    hooks:
+        desc: |
+            Shell commands or scripts to execute before and after a backup or if an error has occurred.
+            IMPORTANT: All provided commands and scripts are executed with user permissions of borgmatic.
+            Do not forget to set secure permissions on this file as well as on any script listed (chmod 0600) to
```
witten: Should the scripts themselves be 0700? Or will `check_call()` execute them regardless of whether they're marked executable?

borgmatic/config/schema.yaml
```diff
@@ -157,3 +157,52 @@ map:
                 desc: Restrict the number of checked archives to the last n. Applies only to the
                       "archives" check.
                 example: 3
+    hooks:
+        desc: |
+            Shell commands or scripts to execute before and after a backup or if an error has occurred.
+            IMPORTANT: All provided commands and scripts are executed with user permissions of borgmatic.
```
witten: Excellent point! Thanks for thinking through (and writing down) the security aspects of this.

borgmatic/config/validate.py
```diff
@@ -44,11 +44,7 @@ def parse_configuration(config_filename, schema_filename):
     except yaml.error.YAMLError as error:
         raise Validation_error(config_filename, (str(error),))
 
-    # pykwalify gets angry if the example field is not a string. So rather than bend to its will,
-    # simply remove all examples before passing the schema to pykwalify.
-    for section_name, section_schema in schema['map'].items():
-        for field_name, field_schema in section_schema['map'].items():
-            field_schema.pop('example')
+    clean_pykwalify(schema)
```
witten: Good call factoring this out.

borgmatic/config/validate.py
```diff
@@ -59,6 +55,17 @@ def parse_configuration(config_filename, schema_filename):
...
 
+def clean_pykwalify(schema):
+    '''
+    pykwalify gets angry if the example field is not a string. So rather than bend to its will,
+    simply remove all examples before passing the schema to pykwalify.
+    '''
+    for section_name, section_schema in schema['map'].items():
+        section_schema.pop('example', None)
+        if 'map' in section_schema:
+            clean_pykwalify(section_schema)
```
witten: I recommend adding tests for this function. And perhaps mocking it out of any unit tests of its direct caller (`parse_configuration()`).

Alternately, if you lose some of the schema nesting as suggested above, you may not actually need this change right now.

borgmatic/tests/unit/borg/test_hook.py
```diff
@@ -0,0 +1,10 @@
+from flexmock import flexmock
+
+from borgmatic.commands import hook as module
+
+
+def test_exec_cmd():
```
witten: Even though there's only a single test case, it'd be helpful to encode in the test name what exactly you're testing about the unit under test. Example: `test_exec_cmd_invokes_each_command()`.

Speaking of which, either this test case or another test case could test that it actually calls multiple different commands if there are indeed multiple in the list. Not strictly necessary though.

borgmatic/commands/hook.py
```diff
@@ -0,0 +1,7 @@
+import subprocess
+
+
+def exec_cmd(config):
```
witten: Even though I don't have a formal coding style convention written down in this project, I generally try to err on the side of verbosity rather than terseness in variable and function naming. So perhaps expand this out to `execute_command()` or similar. Or maybe even better: `execute_hook()`, because that's more specific to the domain in question (hooks). And in fact there may be multiple commands executed by a single hook!

Churro:
Thank you for looking this through!
I will happily provide another commit once we have sorted out the points above :-)

borgmatic/commands/borgmatic.py
```diff
@@ -104,6 +107,9 @@ def main():  # pragma: no cover
                     )
                 if args.check:
                     check.check_archives(args.verbosity, repository, consistency, remote_path=remote_path)
+
+            hook.exec_cmd(hooks.get('after_backup'))
     except (ValueError, OSError, CalledProcessError) as error:
+        hook.exec_cmd(hooks.get('on_error'))
```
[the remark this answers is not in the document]
Churro: Very good catch! I actually overlooked this...

I'd to propose to add a separate ```try / except``` clause enclosing everythin between the ```before_backup``` and ```after_backup``` hook. This would enable a distinct ```on_error``` hook per config file.

IMHO, accumulating all ```on_error``` hooks would contradict the fact that they can be defined individually for each config file.  Dropping ```on_error``` would probably prevent some use cases, such as getting a targeted notification in case a backup process has failed.

borgmatic/config/schema.yaml
Churro: My bad! 0700 is correct, 0600 would give "permission denied", just as with an ordinary shell.

borgmatic/config/validate.py
Churro: Alright, if not needed by the schema, I also suggest not to change this for now. If needed at a later point in time, it could still be (re-)introduced.

borgmatic/commands/borgmatic.py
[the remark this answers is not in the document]
witten: The `try`/`except` option as you describe sounds good. However, if you don't `raise` there, it would be good to avoid swallowing error information. For instance, without a `raise`, borgmatic could encounter an error, you'd get notified of the error (via an `on_error` script), but then you'd have no further information about what went wrong nor any way to debug the problem.

One way I could see dealing with this is adding some logging of the error information, so the information isn't lost. But that might be blocked on https://tree.taiga.io/project/witten-borgmatic/issue/33 (which I've actually started working on).

Another option would be to `raise` for now, and just say that for the time being, an error in one config file causes borgmatic to immediately exit, which incidentally is the current behavior. We can always revise/improve that in a subsequent release.

borgmatic/commands/hook.py
```diff
@@ -0,0 +1,7 @@
...
+    if config and config.get('enable_hook', None) is True:
```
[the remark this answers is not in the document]
witten: Cool, sounds good.

Side note: You bring up an interesting point that the `generate-borgmatic-config` file isn't really usable as-is. I wonder if it would be helpful to have a way to generate a more minimal file that is closer to usable out of the box.

borgmatic/config/schema.yaml
witten: Although you are correct here that 0600 works for the borgmatic config file itself. That doesn't need to be executable, of course.

borgmatic/config/validate.py
witten: Okay, makes sense.

Churro:
I have now consolidated all points. Please let me know what you think about it 👍

b3vis:
I was just coming here to raise an issue asking for this exact request, having backup notifications with the summary displayed on the console in an email or push notification would be extremely useful.

borgmatic/commands/borgmatic.py
```diff
@@ -84,26 +85,33 @@ def main():  # pragma: no cover
...
+                            args.verbosity,
+                            repository,
+                            location,
+                            storage,
+                        )
+                    if args.check:
+                        check.check_archives(args.verbosity, repository, consistency, remote_path=remote_path)
+
+                hook.execute_hook(hooks.get('after_backup'))
+            except (OSError, CalledProcessError):
```
witten: Given that you're re-raising the exception, you *could* `except:` all errors here. It kind of depends on what you want the contract to be for "on_error" hooks.. Should they trigger on all error types or just these particular ones?

Churro: Catching all errors with ```except:``` would imply that an unhandled error would be raised as well. Since the outer exception handler can only deal with ```ValueError, OSError, CalledProcessError```, unhandled exceptions would still be fatal. Hence, it probably makes sense to keep the inner ```except``` consistent with the outer.

witten: Yup, makes sense. Just didn't know if you wanted fatal exceptions to call the "on_error" hook as well before.. being fatal.

PR #20, opened 2018-01-17
Add `encryption_passcommand` configuration option

newtonne:
Using the `passcommand` option is safer than `passphrase` since it removes the need to store the encryption passphrase in the plain text config files.


witten:
Looks great.. Thanks for doing this! And let me know if you'd like to be included in the AUTHORS file either by name or handle.

newtonne:
Thanks for merging 😊

Sounds good. Please add my handle.

witten:
Done!

PR #21, opened 2018-02-03
Fix issue when using both --dry-run and -v options

newtonne:
Whilst trying to do a backup with both `--dry-run` and verbosity (either `-v 1` or `-v 2`) set, I encountered the following error:

    borg create: error: argument -n/--dry-run: not allowed with argument -s/--stats

This pull request is a pretty quick and dirty fix.
Should probably also add some unit tests that have both verbosity and dry-run set.

### DIFF
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -123,8 +123,8 @@ def create_archive(
     files_cache_flags = ('--files-cache', files_cache) if files_cache else ()
     remote_path_flags = ('--remote-path', remote_path) if remote_path else ()
     verbosity_flags = {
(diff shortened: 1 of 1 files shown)


witten:
Thanks for fixing this! I'll go ahead and add some tests separately.

By the way, this issue does not appear to manifest in Borg 1.1.2, but does in Borg 1.1.4. So it must've been something introduced relatively recently.

newtonne:
It seems --dry-run was made incompatible with --stats in Borg 1.1.3. See [Issue #3298](https://github.com/borgbackup/borg/issues/3298).

witten:
This has been released as part of borgmatic 1.1.15!

PR #22, opened 2018-03-03
Only check archives with matching prefix

nickw444:
Borg supports checking a subset of archives in a repo by providing a prefix `--prefix PREFIX` when doing `check`. This PR adds functionality to use the same prefix defined in `retention` config to be passed as this param to borg.

http://borgbackup.readthedocs.io/en/stable/usage/check.html

### DIFF
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -136,7 +136,6 @@ def run_configuration(config_filename, args):  # pragma: no cover
                     repository,
                     storage,
                     consistency,
-                    retention,
                     local_path=local_path,
                     remote_path=remote_path,
                 )
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -136,6 +136,7 @@ def run_configuration(config_filename, args):  # pragma: no cover
                     repository,
                     storage,
                     consistency,
+                    retention,
                     local_path=local_path,
                     remote_path=remote_path,
                 )
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -218,6 +218,13 @@ map:
                 desc: Restrict the number of checked archives to the last n. Applies only to the
                       "archives" check.
                 example: 3
+            archive_prefix:
+                type: scalar
+                desc: |
+                    When performing consistency checks, only consider archive names starting with
+                    this prefix. Borg placeholders can be used. See the output of
+                    "borg help placeholders" for details. Default is "{hostname}-".
+                example: sourcehostname
     hooks:
         desc: |
             Shell commands or scripts to execute before and after a backup or if an error has occurred.
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -218,7 +218,7 @@ map:
                 desc: Restrict the number of checked archives to the last n. Applies only to the
                       "archives" check.
                 example: 3
-            archive_prefix:
+            prefix:
                 type: scalar
                 desc: |
                     When performing consistency checks, only consider archive names starting with
--- a/borgmatic/config/validate.py
+++ b/borgmatic/config/validate.py
@@ -50,11 +50,12 @@ def apply_logical_validation(config_filename, parsed_configuration):
             )
         )
 
-    consistency_archive_prefix = parsed_configuration.get('consistency', {}).get('archive_prefix')
-    if archive_name_format and not consistency_archive_prefix:
+    consistency_prefix = parsed_configuration.get('consistency', {}).get('prefix')
+    if archive_name_format and not consistency_prefix:
         raise Validation_error(
             config_filename, (
-                'If you provide an archive_name_format, you must also specify a consistency archive_prefix.',
+                'Since version 1.2.0, if you provide `archive_name_format`, you must also specify'
+                ' `consistency.prefix`.',
             )
         )
 
--- a/borgmatic/config/validate.py
+++ b/borgmatic/config/validate.py
@@ -8,6 +8,8 @@ import pykwalify.errors
 from ruamel import yaml
 
 
+logger = logging.getLogger(__name__)
+
 def schema_filename():
     '''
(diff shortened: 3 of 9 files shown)


witten:
Thanks for doing this, and for adding a test! Seems like a useful thing to pass through to Borg. What do you think of adding a new `prefix` option to borgmatic's `consistency` config section, rather than reusing the existing `retention` `prefix` option for this consistency check? While it's perhaps a little obnoxious to have multiple prefix options, it would better segregate things so `retention` only deals with pruning, and `consistency` only deals with checks.

witten:
Also, feel free to add yourself to AUTHORS, or just let me know how you'd like attribution.

nickw444:
Sounds good, i'll make those changes and add it to the schema definition!

borgmatic/tests/unit/borg/test_check.py
```diff
@@ -89,7 +89,7 @@ def test_make_check_flags_with_default_checks_and_last_returns_last_flag():
 )
 def test_check_archives_calls_borg_with_parameters(checks):
     check_last = flexmock()
-    consistency_config = flexmock().should_receive('get').and_return(check_last).mock
+    consistency_config = {'check_last': check_last}
```
nickw444: I removed the flexmock usage here. It felt as though it was not adding value. The method being tested should transform config into an output, it doesn't make sense for the test to be concerned about what keys the _black box_ method is using from the config object passed

borgmatic/tests/unit/config/test_validate.py
```diff
@@ -23,13 +23,37 @@ def test_apply_logical_validation_raises_if_archive_name_format_present_without_
             },
         )
 
+def test_apply_logical_validation_raises_if_archive_name_format_present_without_retention_prefix():
```
nickw444: This is going to be a breaking change for existing configurations. Users who previously used `archive_name_format` will need to update their config to add the new `consistency.archive_prefix` option.

How do you handle configuration schema compatibility?

borgmatic/config/validate.py
```diff
@@ -50,6 +50,14 @@ def apply_logical_validation(config_filename, parsed_configuration):
...
         )
 
+    consistency_archive_prefix = parsed_configuration.get('consistency', {}).get('archive_prefix')
+    if archive_name_format and not consistency_archive_prefix:
+        raise Validation_error(
+            config_filename, (
+                'If you provide an archive_name_format, you must also specify a consistency archive_prefix.',
+            )
+        )
+
```
witten: You could in theory omit this validation for the consistency `archive_prefix` value. The check exists for the retention `prefix` so that you don't actually prune archives that you aren't creating as part of your config (potentially resulting in data loss). But unlike pruning, it wouldn't be a huge deal if you accidentally do some extra consistency checks.

I could also see an argument for leaving it in. Let me know which you'd prefer.

borgmatic/config/schema.yaml
```diff
@@ -218,6 +218,13 @@ map:
                 desc: Restrict the number of checked archives to the last n. Applies only to the
                       "archives" check.
                 example: 3
+            archive_prefix:
```
witten: You could just call this `prefix` for parity with the `retention` section's `prefix` and also Borg's own flag. Do not feel strongly either way.

borgmatic/commands/borgmatic.py
```diff
@@ -136,6 +136,7 @@ def run_configuration(config_filename, args):  # pragma: no cover
                     repository,
                     storage,
                     consistency,
+                    retention,
```
witten: This argument not needed any longer?

borgmatic/tests/unit/borg/test_check.py
witten: Yup, makes sense. Thanks!

borgmatic/tests/unit/config/test_validate.py
witten: I generally don't! I've only one had one majorly breaking change, and that was the transition from an INI-like config file to YAML. And for that I included (and required using) a transition tool.

This is more minor, so it might be okay to just ship it and include a decent error message. See my separate comment though about whether this validation is even necessary.

borgmatic/config/validate.py
nickw444: Well speaking from experience (I introduced this tool into my stack last night), I only realised I needed to set the retention `prefix` value because of the validation error message. 

I feel it would be valuable to new users to have this error be raised. 

As for existing users, however, I imagine there would be complaints (or issues raised) where they upgrade `borgmatic` and find their exiting configs no longer work. 

As you mentioned, perhaps we could have some better messaging around when and why this was introduced, so that existing users getting the error can change their configuration accordingly. 

Maybe something along the lines of:

```
Since version 1.2.0, if you provide `archive_name_format`, you must also specify `consistency.prefix`.
```

I noticed you use semver, so maybe this is acceptable?

borgmatic/config/schema.yaml
nickw444: Sounds good!

borgmatic/config/validate.py
witten: Sure, that would work. I honestly expect users to read the release notes when upgrading (maybe that's expecting too much!), so a minor incompatibility like that is probably fine.

There is an alternative, which is just making this a warning rather than an outright error if you want the user to be notified without considering this fatal. There's no precedent for that right now, but it would be pretty easy to just print to stderr, for instance.

borgmatic/tests/unit/config/test_validate.py
nickw444: I've taken the warn approach. I've modified the logging configuration to default to "Warn" level, instead of error. 

I think this should give enough visibility to the change.

nickw444:
Changes up! let me know if there's anything else 👍

nickw444:
Thanks for the speedy review! Fastest turnaround I've had on any open source project. 

(And thanks for making this awesome tool!)

witten:
You caught me at a good time. :smile:

PR #23, opened 2018-08-09
Fixed ruamel version (not working with ruamel >=0.15.5)

filicivi:
Ruamel 0.15.5 (released about 13 hours ago) not working correctly in conjunction with pykwalify.
With a default config, when borgmatic starts 

_pykwalify.errors.RuleError: <RuleError: error code 4: Value for keyword 'map/mapping' is not a dict: Path: '/'>_

Installing ruamel.yaml==0.15.4 seem to solve the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
     ],
     install_requires=(
         'pykwalify>=1.6.0',
-        'ruamel.yaml>0.15.0,<0.16.0',
+        'ruamel.yaml>0.15.0,<=0.15.4',
         'setuptools',
     ),
     tests_require=(
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -2,4 +2,4 @@ flexmock==0.10.2
 pykwalify==1.6.1
 pytest==3.6.3
 pytest-cov==2.5.1
-ruamel.yaml>0.15.0,<0.16.0
+ruamel.yaml>0.15.0,<=0.15.4


witten:
Thanks for the PR. Looks like this issue was introduced by `ruamel.yaml` 0.15.52 specifically. `ruamel.yaml` 0.15.51 does not appear to exhibit this problem. 

In doing some initial diagnosing with borgmatic, I looked at the type information of the `ruamel.yaml`-loaded schema before it's passed to pykwalify. With 0.15.51, it's a `CommentedMap`. With 0.15.52, it's an `ordereddict`. That's probably what's throwing off pykwalify.

It's unclear based on the [changelog](https://pypi.org/project/ruamel.yaml/#changelog) what might have caused this breakage in 0.15.52, but both `CommentedMap` and `ordereddict` are mentioned there.

I want to look at this a little more before just pinning to an older version as you suggest here.

BarbzYHOOL:
Using borgmatic 1.2.0 would solve this issue?

witten:
@BarbzYHOOL borgmatic does not declare a pinned dependency on a particular version of borgbackup. Generally most versions of borgbackup should work with most versions of borgmatic.

Having said that, I plan to rectify this issue soon, so you won't need to pin borgmatic anymore.

witten:
Okay, I believe this is now fixed in borgmatic 1.2.2, just released. Rather than pinning `ruamel.yaml` as in this PR, I managed to change borgmatic's use of `ruamel.yaml` to load the config and the schema in such a way that pykwalify no longer explodes. I confirmed that the changed code works with both `ruamel.yaml` 0.15.51 and 0.15.52 (before and after the breaking change, respectively).

I do appreciate the PR though. It let me know this was an issue, and it would've been the backup plan had I not been able to work around the issue.

witten:
Also, do you recall what the incompatibility was with Borg 1.0.9? Was it with the --info sub-command argument ordering, or something else? A traceback/log would be super helpful.

witten:
Okay, I'll see if I can reproduce the problem on my machine. If this sort of thing occurs again, a bug report would be great. Thanks.

BarbzYHOOL:
Will do but well I needed to test borgmatic first before opening a bug report :P

n3storm:
> Okay, I believe this is now fixed in borgmatic 1.2.2, just released. Rather than pinning `ruamel.yaml` as in this PR, I managed to change borgmatic's use of `ruamel.yaml` to load the config and the schema in such a way that pykwalify no longer explodes. I confirmed that the changed code works with both `ruamel.yaml` 0.15.51 and 0.15.52 (before and after the breaking change, respectively).
> 
> I do appreciate the PR though. It let me know this was an issue, and it would've been the backup plan had I not been able to work around the issue.

Just installed borgmattic and got this error.

```
> pip3 show borgmatic
> Name: borgmatic
> Version: 1.2.2
> Summary: A wrapper script for Borg backup software that creates and prunes backups
> Home-page: https://torsion.org/borgmatic
> Author: Dan Helfman
> Author-email: witten@torsion.org
> License: UNKNOWN
> Location: /usr/local/lib/python3.5/dist-packages
> Requires: setuptools, ruamel.yaml, pykwalify
```

n3storm:
This happens with ruamel.yaml 0.15.70. Moving to 0.15.69 solved this issue.

witten:
Thanks for mentioning. This issues is being tracked here: https://projects.torsion.org/witten/borgmatic/issues/95 .. You're welcome to follow its progress there!

n3storm:
sorry, thanks!

witten:
No worries! Looks like the issue is now fixed in ruamel.yaml 0.15.71.

PR #24, opened 2018-09-04
fixed error: /bin/ash: root: not found

grerrg:
This example is better suited for docker images based on alpine-linux. although just specifying the executable `borgmatic` is more general purpose i think. like this:
`0 3 * * * borgmatic 2>&1`

The output would show up in `docker logs` if you would run this an a container. Also in the syslog in any other scenario.

### DIFF
--- a/sample/cron-alpine/borgmatic
+++ b/sample/cron-alpine/borgmatic
@@ -1,3 +1,3 @@
 # You can drop this file into /etc/cron.d/ to run borgmatic nightly.
 
-0 3 * * * root PATH=$PATH:/usr/local/bin /usr/local/bin/borgmatic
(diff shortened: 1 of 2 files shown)


witten:
Thanks for submitting! The existing example (with user `root` and the current path) works on non-Docker images like a standard Ubuntu system, so I'd prefer to keep that example working if possible. But I think you have a point about running with Alpine. What do you think of making separate files for each of the examples, so we can demonstrate both?

Sample path idea: `sample/cron-alpine/borgmatic` or similar.

grerrg:
I'll go with it. Having more info given is always good.

witten:
Cool, thank you!

PR #27, opened 2018-09-27
check_archives does not take json parameter

irrenhaus:
Borgmatic crashed on my machine because check_archives does not expect a parameter `json`.


witten:
Thank you for catching this, and especially for the fix!

@floli FYI.

witten:
Just released this fix as borgmatic 1.2.4.

irrenhaus:
Now, that was a quick reaction! Thank you very much!

PR #29, opened 2019-01-21
Add link to  openSUSE packages

antlarr:
Add a link to the software.opensuse.org page were both official and community packages of borgmatic are available to be downloaded or installed using 1-click-install.


witten:
Cool, thank you for pointing this out!

PR #30, opened 2019-05-20
Unlike borgmatic, borg expects an initital command without hyphens

Steve-V:
Minor formatting cleanup, might throw off a user new to borg/borgmatic

### DIFF
--- a/docs/how-to/set-up-backups.md
+++ b/docs/how-to/set-up-backups.md
@@ -120,7 +120,7 @@ this step if you already have a Borg repository.) To create a repository, run
 a command like the following:
 
 ```bash
-borgmatic --init --encryption repokey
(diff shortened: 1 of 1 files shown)


witten:
Hey Steve-V, thanks for the PR! However this looks like this change is to the borgmatic command-line, which *does* actually expect `--init` rather than `init`. This is for borgmatic's built-in action that effectively wraps `borg init` for convenience (reading repository names from borgmatic's config, etc.).

Steve-V:
Whoops....apologies, you're exactly right.  Thanks for the fast response and for such useful software!

witten:
Not a problem! If you find any other potential documentation or other bugs, please don't hesitate to submit another PR.

PR #32, opened 2019-12-13
Use --remote-path, --debug and --info when checking for repo existence

megabug:
User-defined remote paths and verbosity settings are not being honoured in the first call to borg in borgmatic's init command. This call is used to check whether the repo already exists or not.

I've noticed this while creating a new repo and trying to use a remote-path set to force the use of a newer version of borg on a remote host instead of an old version, leading to scary error messages from an ancient version of borg during repo init.


witten:
Good find! This looks right to me, and it doesn't look like any tests need updating as a result of this change. Thanks for the PR.

witten:
This fix is released as part of borgmatic 1.4.20. Thanks again!

PR #33, opened 2020-01-19
Support directory format dump cleanup.

rdunklau:
Previously, only deleting a dump in a single-file format was supported.
This led to errors when performing a PostgreSQL directory format backup.


witten:
Okay, this is now released in borgmatic 1.4.22! Thanks again.

PR #34, opened 2020-04-10
Fixed Broken pg_dumpall Restore

ndbeals:
When the database name is set to 'all', borgmatic uses `pg_dumpall` instead of `pg_dump` to dump the database. `pg_dumpall` outputs a plain-text format and `pg_restore` does not accept that format. To run these plain-text outputs, `psql` is needed instead.

This pull request implements that. Thanks!

### DIFF
--- a/borgmatic/hooks/postgresql.py
+++ b/borgmatic/hooks/postgresql.py
@@ -34,7 +34,7 @@ def dump_databases(databases, log_prefix, location_config, dry_run):
         )
         all_databases = bool(name == 'all')
         command = (
-            ('pg_dumpall' if all_databases else 'pg_dump', '--no-password', '--clean')
+            ('pg_dumpall' if all_databases else 'pg_dump', '--no-password', '--clean','--if-exists')
             + ('--file', dump_filename)
             + (('--host', database['hostname']) if 'hostname' in database else ())
             + (('--port', str(database['port'])) if 'port' in database else ())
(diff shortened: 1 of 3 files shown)


witten:
Thank you so much for finding and fixing this! Would you like to take a stab at updating tests for this fix? If so, here are some places to start:

 * In `tests/end-to-end/test_database.py`, in the test borgmatic configuration, you could add a `- name: all` config to the test databases. That should exercise the new code path you added for "all" database dumps/restores.
 * Some of the unit tests will need updating based on the new command arguments. They can be found in `tests/unit/hooks/test_postgresql.py`.

Either way, let me know!

ndbeals:
I finished those tests, including adding another unit test. It passes all tests, including end-to-end.

There is a caveat to dumping all on a postgres database. A dump all includes all users, including the one you log into the database as to dump it. When restoring the database, even a brand-new postgres database has a user (defaults as 'postgres') which is required to run the restore all through. In short this means the user will these errors show up in their console:
```
ERROR:  current user cannot be dropped
STATEMENT:  DROP ROLE IF EXISTS postgres;
ERROR:  role "postgres" already exists
STATEMENT:  CREATE ROLE postgres;
``` 

But that's okay, It's failing successfully. Any further errors they get may be a problem though. It's a quirk of postgres and I don't see how it can be handled in borgmatic at all.

witten:
Awesome.. Thanks for taking the time to add tests! And for the tip about the postgres users.. I'm glad to hear that's "just" a warning and not a blocker.

witten:
Just released in borgmatic 1.5.2! Thanks again.

ndbeals:
Glad to have helped! you made it easy.

PR #35, opened 2020-08-07
Doc config after_everything is "after" not before

jduchateau:
Found a small mistake in the `after_everything` comment in the configuration file.


witten:
Awesome, thank you for taking the time to make this fix!

PR #38, opened 2021-03-19
Fix link to issue tracker in documentation

lukehsiao:
Fixes: a1d986d952ac739b52e11e755e834dcf7444d957

### DIFF
--- a/docs/_includes/components/suggestion-link.html
+++ b/docs/_includes/components/suggestion-link.html
@@ -1,7 +1,7 @@
 <h2>Improve this documentation</h2>
 
 <p>Have an idea on how to make this documentation even better? Use our <a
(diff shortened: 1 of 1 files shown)


witten:
This link was actually intentional, in that I was trying to give users a little overview of the support options, which itself includes a link to the issue tracker. But I'm totally fine changing it if you think a direct link is better!

witten:
Thanks for the PR!

lukehsiao:
That makes sense, too. If that's the case, I'd suggest maybe linking to https://torsion.org/borgmatic/#support-and-contributing and saying something like

> See how you can make contributions here!

rather than

> Use our issue tracker to send your feedback!

Or something along those lines?

PR #41, opened 2021-07-27
tests/integration/test_execute: use plain Python rather than xxd

mkszuba:
This allows test_execute not to depend on either vim or /dev/urandom.


witten:
Excellent idea! I had no idea xxd came from vim.

PR #43, opened 2021-11-25
Github doesn't allow script embedding

acsfer:
For acsiinema screencast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ hooks:
 Want to see borgmatic in action? Check out the <a
 href="https://asciinema.org/a/203761" target="_blank">screencast</a>.
 
-[![asciicast](https://asciinema.org/a/203761.png)](https://asciinema.org/a/203761)
+<a href="https://asciinema.org/a/203761"><img src="https://asciinema.org/a/203761.png" width="480"></a>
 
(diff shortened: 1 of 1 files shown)


witten:
Thanks for taking the time to submit this! Here's the difficulty I'm seeing though.. borgmatic is primarily hosted on [torsion.org](https://torsion.org/borgmatic/) and mirrored on GitHub for convenience. On torsion.org, the `<script>` in the README works just fine today and actually embeds the asciinema player. And switching to the `[asciicast]` Markdown tag as you have here would be cool with me if it makes the experience better for GitHub users. However, on torsion.org, the `[asciicast]` tag switches it from an embedded player to an image link (okay, that could still be tolerable), but then makes the image *humongous* as it's not respecting the containing div's width for some reason. And I can't see a way to set the width in Markdown since it's no longer using an HTML-style tag.

If you want to play around with this locally, you can run `scripts/dev-docs` and then point your browser at http://localhost:8080 to see what's generated.

Let me know your thoughts!

acsfer:
Oh I see, and got your points. I've switched to HTML :)
Feel free to close if you prefer keep things as is.

witten:
This seems like a reasonable compromise! Thanks again for doing this.

PR #45, opened 2022-07-13
Added link to borgbackup list --format docs

Jelle-SamsonIT:
I kept searching for this link, so it's time to add it to official docs.


witten:
Thanks for taking the time to submit this PR! I'm always looking for ways to improve the docs. Do you think this Borg format link would be better to link to though? https://borgbackup.readthedocs.io/en/stable/usage/list.html#the-format-specifier-syntax

witten:
Awesome, thank you!

PR #46, opened 2022-07-13
Enhanced docs with info on fetching mysql database size

Jelle-SamsonIT:
Not sure this is the final wording yet, but I think this is handy information for newcomers.


witten:
I apologize for the lengthy delay in getting to this. I'll merge it and then generalize the text so it applies to the various database types supported by borgmatic rather than just MySQL. I may also move it to the database page.

PR #48, opened 2022-07-23
feat: add verify_ssl flag for Healthchecks

Tooa:
Hi :wave:, 

the following pull request adds the optional flag `verify_ssl` to the Healthchecks monitoring configuration. The variable defaults to `True`.

**Use case:** When running a self-hosted Healthchecks instance with self-signed certificates the ping command fails with the error:

```bash
/etc/borgmatic.d/config.yaml: Pinging Healthchecks finish
/etc/borgmatic.d/config.yaml: Healthchecks error: HTTPSConnectionPool(host='<local_network_dns_redacted>', port=443): Max retries exceeded with url: /ping/1bb6a6f7-679f-4b14-8f47-4480b542a28a (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)')))
``` 

Let me know what you think. Hope you don't mind me creating the pull request here at Github rather than over at torsion.org.

### DIFF
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -1012,6 +1012,12 @@ properties:
                             Healthchecks ping URL or UUID to notify when a
                             backup begins, ends, or errors.
                         example: https://hc-ping.com/your-uuid-here
+                    verify_tls:
+                        type: boolean
+                        description: |
+                            Verify the SSL certificate of the endpoint.
+                            Defaults to true.
+                        example: false
                     send_logs:
                         type: boolean
                         description: |
--- a/borgmatic/hooks/healthchecks.py
+++ b/borgmatic/hooks/healthchecks.py
@@ -125,7 +125,9 @@ def ping_monitor(hook_config, config_filename, state, monitoring_log_level, dry_
     if not dry_run:
         logging.getLogger('urllib3').setLevel(logging.ERROR)
         try:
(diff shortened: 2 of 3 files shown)


witten:
Thank you so much for taking the time to develop and submit this.. with tests no less! Would you mind renaming `verify_ssl` to `verify_tls` though, since these days we really [use TLS everywhere instead SSL](https://kinsta.com/knowledgebase/tls-vs-ssl/)? Small thing, I know, but I'm kind of nerdy like that.

Of course, you are probably aware that by disabling TLS verification like this, you aren't getting (all of) the benefits of encryption when talking to your self-hosted Healthchecks server.

And creating the PR on GitHub is totally fine. That's why I have PRs open here, because I realize it's more convenient for some folks.

Thanks!

Tooa:
> Would you mind renaming verify_ssl to verify_tls though, since these days we really [use TLS everywhere instead SSL](https://kinsta.com/knowledgebase/tls-vs-ssl/)? Small thing, I know, but I'm kind of nerdy like that.

Thank you for your kind feedback. I have adapted the code accordingly. Let me know if you want something else changed.

witten:
I don't mind the warning if you don't, as it's technically correct and lets the user know that they are running without a net, so to speak. I'll merge this now!

witten:
Thanks again! This will be part of the next release.

PR #49, opened 2022-11-06
specify pg dump/restore commands (#311)

jpaniagualaconich:
add `pg_dump_command`, `pg_restore_command` and `psql_command` configuration options. In this way, users can specify alternative commands on a per-database basis.

This is convenient if one needs to do backup/restore of different versions of postgres. A concrete example is the use of different versions of postgres running in docker containers for which one would want to use commands inside these containers while still being able to use pipes. Example:
```yaml
hooks:
  postgresql_databases:
    - name: my_pg_database
      pg_dump_command: docker exec my_pg_database pg_dump
      pg_restore_command: docker exec my_pg_database pg_restore
      psql_command: docker exec my_pg_database psql
```

It partially fixes (#311) since it's just for the postgres hook.

### DIFF
--- a/borgmatic/hooks/postgresql.py
+++ b/borgmatic/hooks/postgresql.py
@@ -56,9 +56,11 @@ def dump_databases(databases, log_prefix, location_config, dry_run):
         )
         all_databases = bool(name == 'all')
         dump_format = database.get('format', 'custom')
+        default_dump_command = 'pg_dumpall' if all_databases else 'pg_dump'
+        dump_command = database.get('pg_dump_command') or default_dump_command
         command = (
             (
-                'pg_dumpall' if all_databases else 'pg_dump',
+                dump_command,
                 '--no-password',
                 '--clean',
                 '--if-exists',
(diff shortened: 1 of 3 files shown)


borgmatic/hooks/postgresql.py
```diff
@@ -56,9 +56,11 @@ def dump_databases(databases, log_prefix, location_config, dry_run):
         )
         all_databases = bool(name == 'all')
         dump_format = database.get('format', 'custom')
+        default_dump_command = 'pg_dumpall' if all_databases else 'pg_dump'
+        dump_command = database.get('pg_dump_command') or default_dump_command
         command = (
             (
-                'pg_dumpall' if all_databases else 'pg_dump',
+                dump_command,
```
witten, line 63: The only problem I see with this is that the user won't be able to specify separate `pg_dump` and `pg_dumpall` commands. So you could in theory configure an "all" database hook with no way to dump it. What do you think of having a separate `pg_dumpall_command` configuration option to handle this case?

witten (commented):
Thank you so much for taking the time to code this up and add tests, too! This looks like a great addition to have.

borgmatic/hooks/postgresql.py
jpaniagualaconich: hey there! it started like that, with `pg_dump_command` and `pg_dumpall_command`, but then I realised they're never needed for the same `postgresql_databases` item. Using one or the other depends on having `name: my_database` or `name: all`.

With that in mind I went with different semantics: "I don't want borgmatic to choose depending on the value under `name:` and instead I want to provide my own command".

Other possibility that I considered was a _total override_: "let the user provide the exact command line plus switches, with placeholders for name, user, host, port, ..."
It's more verbose which is why I didn't go with that one, but that's just because I'm lazy.

Am I missing something? What do you think?

witten:
I think the rationale you've given for a single `pg_dump_command` makes sense to me. Users _can_ shoot themselves in the foot, but maybe that's okay. Any configuration option that allows a user to override a command can be used incorrectly—not just these ones—so I'm willing to proceed if you are!

jpaniagualaconich:
Well, thank you!

PR #50, opened 2023-03-02
feat: add dump-restore support for sqlite databases

diivi:
fixes [#295](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/295)
Will add to e2e tests once I understand how to do so.

### DIFF
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -938,24 +938,24 @@ properties:
                     required: ['path','name']
                     additionalProperties: false
                     properties:
+                        name:
+                            type: string
+                            description: |
+                                This is used to tag the database dump file 
+                                with a name. It is not the path to the database 
+                                file itself. The name "all" has no special 
+                                meaning for SQLite databases.
+                            example: users
                         path:
                             type: string
                             description: |
                                 Path to the SQLite database file to dump. If
                                 relative, it is relative to the current working
-                                directory. If absolute, it is relative to the
-                                root of the filesystem. Note that using this
+                                directory. Note that using this
                                 database hook implicitly enables both
                                 read_special and one_file_system (see above) to
                                 support dump and restore streaming.
                             example: /var/lib/sqlite/users.db
-                        name:
-                            type: string
-                            description: |
-                                This is used to tag the database dump file with
-                                a name. It is not used to identify the database
-                                file itself.
-                            example: users
             mongodb_databases:
                 type: array
                 items:
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -931,6 +931,31 @@ properties:
                     mysqldump/mysql commands (from either MySQL or MariaDB). See
                     https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html or
                     https://mariadb.com/kb/en/library/mysqldump/ for details.
+            sqlite_databases:
+                type: array
+                items:
+                    type: object
+                    required: ['path','name']
+                    additionalProperties: false
+                    properties:
+                        path:
+                            type: string
+                            description: |
+                                Path to the SQLite database file to dump. If
+                                relative, it is relative to the current working
+                                directory. If absolute, it is relative to the
+                                root of the filesystem. Note that using this
+                                database hook implicitly enables both
+                                read_special and one_file_system (see above) to
+                                support dump and restore streaming.
+                            example: /var/lib/sqlite/users.db
+                        name:
+                            type: string
+                            description: |
+                                This is used to tag the database dump file with
+                                a name. It is not used to identify the database
+                                file itself.
+                            example: users
             mongodb_databases:
                 type: array
                 items:
--- a/borgmatic/hooks/sqlite.py
+++ b/borgmatic/hooks/sqlite.py
@@ -105,7 +105,7 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         return
 
     try:
-        os.remove('/home/divyansh/Desktop/hello.txt')
+        os.remove(database_path)
         logger.warn(f'{log_prefix}: Removed existing SQLite database at {database_path}')
     except FileNotFoundError:  # pragma: no cover
         pass
--- a/borgmatic/hooks/sqlite.py
+++ b/borgmatic/hooks/sqlite.py
@@ -105,7 +105,8 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         return
 
     try:
-        os.remove(database_path)
+        os.remove('/home/divyansh/Desktop/hello.txt')
+        logger.warn(f'{log_prefix}: Removed existing SQLite database at {database_path}')
     except FileNotFoundError:  # pragma: no cover
         pass
 
--- a/borgmatic/hooks/sqlite.py
+++ b/borgmatic/hooks/sqlite.py
@@ -1,6 +1,5 @@
 import logging
 import os
-from subprocess import CalledProcessError
 
 from borgmatic.execute import execute_command, execute_command_with_processes
 from borgmatic.hooks import dump
@@ -29,11 +28,16 @@ def dump_databases(databases, log_prefix, location_config, dry_run):
 
     logger.info('{}: Dumping SQLite databases{}'.format(log_prefix, dry_run_label))
 
-    if databases[0]['name'] == 'all':
-        logger.warning('The "all" database name has no meaning for SQLite3 databases')
-
     for database in databases:
         database_path = database['path']
+
+        if database['name'] == 'all':
+            logger.warning('The "all" database name has no meaning for SQLite3 databases')
+        if not os.path.exists(database_path):
+            logger.warning(
+                f'{log_prefix}: No SQLite database at {database_path}; An empty database will be created and dumped'
+            )
+
         dump_path = make_dump_path(location_config)
         dump_filename = dump.make_database_dump_filename(dump_path, database['name'])
         if os.path.exists(dump_filename):
@@ -101,9 +105,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         return
 
     try:
-        execute_command(('rm', database_path), shell=True)
-    except CalledProcessError:  # pragma: no cover
-        logger.info(f'{log_prefix}: Database does not exist at {database_path}, skipping removal')
+        os.remove(database_path)
+    except FileNotFoundError:  # pragma: no cover
+        pass
 
     restore_command = (
         'sqlite3',
--- a/borgmatic/hooks/sqlite.py
+++ b/borgmatic/hooks/sqlite.py
@@ -17,7 +17,7 @@ def make_dump_path(location_config):  # pragma: no cover
     )
 
 
-def dump_databases(databases, log_prefix, location_config, dry_run):  # pragma: no cover
+def dump_databases(databases, log_prefix, location_config, dry_run):
     '''
     Dump the given SQLite3 databases to a file. The databases are supplied as a sequence of
     configuration dicts, as per the configuration schema. Use the given log prefix in any log
@@ -29,11 +29,13 @@ def dump_databases(databases, log_prefix, location_config, dry_run):  # pragma: 
 
     logger.info('{}: Dumping SQLite databases{}'.format(log_prefix, dry_run_label))
 
+    if databases[0]['name'] == 'all':
+        logger.warning('The "all" database name has no meaning for SQLite3 databases')
+
     for database in databases:
         database_path = database['path']
-        database_filename = database['name']
         dump_path = make_dump_path(location_config)
-        dump_filename = dump.make_database_dump_filename(dump_path, database_filename)
+        dump_filename = dump.make_database_dump_filename(dump_path, database['name'])
         if os.path.exists(dump_filename):
             logger.warning(
                 f'{log_prefix}: Skipping duplicate dump of SQLite database at {database_path} to {dump_filename}'
@@ -98,13 +100,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     if dry_run:
         return
 
-    remove_command = (
-        'rm',
-        database_path,
-    )
     try:
-        execute_command(remove_command, shell=True)
-    except CalledProcessError:
+        execute_command(('rm', database_path), shell=True)
+    except CalledProcessError:  # pragma: no cover
         logger.info(f'{log_prefix}: Database does not exist at {database_path}, skipping removal')
 
     restore_command = (
--- a/borgmatic/hooks/sqlite.py
+++ b/borgmatic/hooks/sqlite.py
@@ -1,0 +1,122 @@
+import logging
+import os
+from subprocess import CalledProcessError
+
+from borgmatic.execute import execute_command, execute_command_with_processes
+from borgmatic.hooks import dump
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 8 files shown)


docs/how-to/backup-your-databases.md
```diff
@@ -154,7 +161,7 @@ bring back any missing configuration files in order to restore a database.
 
 ## Supported databases
 
-As of now, borgmatic supports PostgreSQL, MySQL/MariaDB, and MongoDB databases
+As of now, borgmatic supports PostgreSQL, MySQL/MariaDB, MongoDB databases and SQLite databases
```
diivi, line 164:
```suggestion
As of now, borgmatic supports PostgreSQL, MySQL/MariaDB, MongoDB and SQLite databases
```
just noticed this.

borgmatic/config/schema.yaml
```diff
@@ -931,6 +931,31 @@ properties:
...
+                    type: object
+                    required: ['path','name']
+                    additionalProperties: false
+                    properties:
+                        path:
+                            type: string
+                            description: |
+                                Path to the SQLite database file to dump. If
+                                relative, it is relative to the current working
+                                directory. If absolute, it is relative to the
```
witten, line 946: I think you can safely omit the "If absolute ..." sentence, as it's strongly implied!

borgmatic/config/schema.yaml
```diff
@@ -931,6 +931,31 @@ properties:
...
+                                root of the filesystem. Note that using this
+                                database hook implicitly enables both
+                                read_special and one_file_system (see above) to
+                                support dump and restore streaming.
+                            example: /var/lib/sqlite/users.db
+                        name:
```
witten, line 952: I'd suggest putting this "name" option before "path" so as to keep to the convention set by the other database hooks above.

borgmatic/config/schema.yaml
```diff
@@ -931,6 +931,31 @@ properties:
...
+                            type: string
+                            description: |
+                                This is used to tag the database dump file with
+                                a name. It is not used to identify the database
+                                file itself.
```
witten, line 957: This second sentence might be clearer if it was instead something like: "It is not the path to the database file itself."

borgmatic/hooks/sqlite.py
```diff
@@ -0,0 +1,122 @@
...
+def make_dump_path(location_config):  # pragma: no cover
+    '''
+    Make the dump path from the given location configuration and the name of this hook.
+    '''
+    return dump.make_database_dump_path(
+        location_config.get('borgmatic_source_directory'), 'sqlite_databases'
+    )
+
+
+def dump_databases(databases, log_prefix, location_config, dry_run):  # pragma: no cover
```
witten, line 20: Please remove the "# pragma: no cover" here. We definitely want test coverage to be tracked for main functions like these ones. Thanks!

borgmatic/hooks/sqlite.py
```diff
@@ -0,0 +1,122 @@
...
+    is a dry run, then don't actually dump anything.
+    '''
+    dry_run_label = ' (dry run; not actually dumping anything)' if dry_run else ''
+    processes = []
+
+    logger.info('{}: Dumping SQLite databases{}'.format(log_prefix, dry_run_label))
+
+    for database in databases:
+        database_path = database['path']
+        database_filename = database['name']
```
witten, line 34: This variable is only used once, so you could cut it and just use `database['name']` directly below. (I don't feel strongly either way.)

borgmatic/hooks/sqlite.py
```diff
@@ -0,0 +1,122 @@
...
+    logger.debug(f'{log_prefix}: Restoring SQLite database at {database_path}{dry_run_label}')
+    if dry_run:
+        return
+
+    remove_command = (
+        'rm',
+        database_path,
+    )
+    try:
+        execute_command(remove_command, shell=True)
```
witten, line 106: You could probably inline the removal command here, given it's so short. (Do not feel strongly.)

docs/how-to/backup-your-databases.md
```diff
@@ -27,6 +27,9 @@ hooks:
         - name: posts
     mongodb_databases:
         - name: messages
+    sqlite_databases:
+        - path: /var/lib/sqlite3/mydb.sqlite
+        - name: mydb
```
witten, line 32: I think the "-" shouldn't be on this line, right? Since the path and the name are part of the same list mapping.

docs/how-to/backup-your-databases.md
```diff
@@ -74,6 +77,9 @@ hooks:
           password: trustsome1
           authentication_database: mongousers
           options: "--ssl"
+    sqlite_databases:
+        - path: /var/lib/sqlite3/mydb.sqlite
+        - name: mydb
```
witten, line 82: Same thing here.

docs/how-to/backup-your-databases.md
```diff
@@ -97,7 +103,8 @@ hooks:
 ```
 
 Note that you may need to use a `username` of the `postgres` superuser for
-this to work with PostgreSQL.
+this to work with PostgreSQL. Also, the `all` database name is not supported
+for SQLite databases.
```
witten, line 107: That's a fair omission! Do you think it's worth documenting what happens if `all` is used for a SQLite database.. Or even coding in an error if the user tries to set a name of `all`?

tests/end-to-end/test_database.py
```diff
@@ -73,6 +73,9 @@ def write_configuration(
           hostname: mongodb
           username: root
           password: test
+    sqlite_databases:
+        - path: /tmp/sqlite_test.db
+        - name: sqlite_test
```
witten, line 78: Remove "-" here too.

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+def test_dump_databases_logs_and_skips_if_dump_already_exists():
+    databases = [{'path': '/path/to/database', 'name': 'database'}]
+
+    flexmock(module).should_receive('make_dump_path').and_return('/path/to/dump')
+    flexmock(module).should_receive('dump.make_database_dump_filename').and_return(
+        '/path/to/dump/database'
+    )
+    flexmock(module.os.path).should_receive('exists').and_return(True)
+    flexmock(logging).should_receive('info')
+    flexmock(logging).should_receive('warning')
```
witten, line 18: IMO you don't need to mock out the logging calls at all, as they're part of the standard library and we don't really care about the side effects. Up to you though.

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+
+    flexmock(module).should_receive('make_dump_path').and_return('/path/to/dump')
+    flexmock(module).should_receive('dump.make_database_dump_filename').and_return(
+        '/path/to/dump/database'
+    )
+    flexmock(module.os.path).should_receive('exists').and_return(False)
+    flexmock(logging).should_receive('info')
+    flexmock(logging).should_receive('warning')
+    flexmock(module).should_receive('dump.create_parent_directory_for_dump')
+    flexmock(module).should_receive('execute_command').and_return('process')
```
witten, line 37: Rather than returning the same string everytime, I'd recommend returning a unique flexmock instance, which is really designed for this sort of use case. One way to do that is something like:

```python
processes = [flexmock(), flexmock()]
flexmock(module).should_receive('execute_command').and_return(processes[0]).and_return(processes[1])

assert module.dump_databases(...) == processes
```

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+    ]
+
+    flexmock(module).should_receive('make_dump_path').and_return('/path/to/dump')
+    flexmock(module).should_receive('dump.make_database_dump_filename').and_return(
+        '/path/to/dump/database'
+    )
+    flexmock(module.os.path).should_receive('exists').and_return(False)
+    flexmock(logging).should_receive('info')
+    flexmock(logging).should_receive('warning')
+    flexmock(module).should_receive('dump.create_parent_directory_for_dump')
```
witten, line 36: Does this actually work? You might need to do this instead:

```python
flexmock(module.dump).should_receive('create_parent_directory_for_dump')
```

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+
```
witten, line 56: Since you're testing that something _doesn't_ happen (as per the test function's name), this might be a good place to do:

```python
flexmock(module).should_receive('execute_command').never()
```

You could also slap a `.never()` onto the `create_parent_directory_for_dump` mocking above.

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+    assert module.dump_databases(databases, 'test.yaml', {}, dry_run=True) == []
+
+
+def test_restore_database_dump_restores_database():
+    database_config = [{'path': '/path/to/database', 'name': 'database'}]
+    extract_process = flexmock(stdout=flexmock())
+
```
witten, line 63: You'll also need to mock out `execute_command()` in order to mock out the removal of the database file prior to restoration.

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,89 @@
...
+    flexmock(module).should_receive('execute_command_with_processes').and_return('process')
```
witten, line 64: This mock doesn't need to return anything.

tests/unit/hooks/test_sqlite.py
witten: An additional test you could add as well in a separate test function: Testing what happens when the database doesn't exist and you skip the removal.

witten (changes requested):
This is a great start! Thank you so much for tackling this ticket. I think this is a much-needed addition to borgmatic's database feature set. Let me know if I can help out at all with the end-to-end tests.

docs/how-to/backup-your-databases.md
diivi: Maybe a message with `logger.info` is enough, letting the user know it's treated as a general name for sqlite databases instead of a special keyword.

witten: Sounds good!

borgmatic/hooks/sqlite.py
```diff
@@ -0,0 +1,120 @@
...
+    entries. Use the given location configuration dict to construct the destination path. If this
+    is a dry run, then don't actually dump anything.
+    '''
+    dry_run_label = ' (dry run; not actually dumping anything)' if dry_run else ''
+    processes = []
+
+    logger.info('{}: Dumping SQLite databases{}'.format(log_prefix, dry_run_label))
+
+    if databases[0]['name'] == 'all':
+        logger.warning('The "all" database name has no meaning for SQLite3 databases')
```
witten, line 33: This looks like it's only checking the first database for the name `all`, when in practice a user *could* configure multiple databases and, for example, name the second or third one `all`. If you wanted to deal with this you could do this check for the current database within the `for` loop below instead of here.

borgmatic/hooks/sqlite.py
```diff
@@ -0,0 +1,120 @@
...
+        raise ValueError('The database configuration value is invalid')
+
+    database_path = database_config[0]['path']
+
+    logger.debug(f'{log_prefix}: Restoring SQLite database at {database_path}{dry_run_label}')
+    if dry_run:
+        return
+
+    try:
+        execute_command(('rm', database_path), shell=True)
```
witten, line 104: It just occurred to me that you could replace this with the Python native `os.remove()` since we don't need anything special provided by `execute_command()`. Up to you though.

If you do make that change, the `except` on the next line would change to a `FileNotFoundError`.

tests/end-to-end/test_database.py
```diff
@@ -73,6 +73,9 @@ def write_configuration(
           hostname: mongodb
           username: root
           password: test
+    sqlite_databases:
+        - name: sqlite_test
+          path: /tmp/sqlite_test.db
```
witten, line 78: Unless I'm missing it, it doesn't look like `/tmp/sqlite_test.db` is actually populated anywhere. So when the test dumps the database file, it's simply dumping.. nothing..? (I just confirmed that `sqlite3` will happily dump a non-existent file.) So do you think it's worth trying to create this as a SQLite test file within the test, so it's actually dumping something?

Related question: Do you think borgmatic should warn or error if the `path` file to be dumped doesn't exist? Or should that be on the user to worry about?

tests/unit/hooks/test_sqlite.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_dump_databases_logs_and_skips_if_dump_already_exists():
+    databases = [{'path': '/path/to/database', 'name': 'database'}]
+
+    flexmock(module).should_receive('make_dump_path').and_return('/path/to/dump')
+    flexmock(module.dump).should_receive('make_database_dump_filename').and_return(
+        '/path/to/dump/database'
+    )
+    flexmock(module.os.path).should_receive('exists').and_return(True)
+
```
witten, line 17: You could do a `.never()` assertion here that no commands are executed (like you do in tests below). Up to you!

witten (changes requested):
A few minor comments, but this is looking really good and I think will be ready for merging soon! Thank you.

tests/end-to-end/test_database.py
diivi: For the first part, I don't think the contents of the file really matter for the test?
Second part, yeah, I value UX and think that a simple log should be done to let the user know that the file didn't exist firsthand, and they are creating an empty database and dumping it's contents.

witten: Okay, I'm fine with dumping nothing if you are! And a warning sounds great.

borgmatic/hooks/sqlite.py
```diff
@@ -101,9 +105,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         return
 
     try:
-        execute_command(('rm', database_path), shell=True)
-    except CalledProcessError:  # pragma: no cover
-        logger.info(f'{log_prefix}: Database does not exist at {database_path}, skipping removal')
+        os.remove(database_path)
```
witten, line 108: IMO this should be mocked out in any test that gets this far, so as to avoid attempting file removal on the filesystem during tests. I realize it's unlikely to delete anything real, but I try to avoid unintended filesystem side-effects in tests. Sorry to be a pain here!

witten (changes requested):
One last thing and then I think we're all done!

borgmatic/hooks/sqlite.py
diivi: No worries! I'm not in a hurry to get this merged 😄. Nice catch though, helps me think more about the side effects of my code.

borgmatic/hooks/sqlite.py
```diff
@@ -105,7 +105,8 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         return
 
     try:
-        os.remove(database_path)
+        os.remove('/home/divyansh/Desktop/hello.txt')
```
witten, line 108: Looks like you've maybe left a test path in here...?

witten (changes requested):

borgmatic/hooks/sqlite.py
diivi: ah, my bad.

witten (approved):
This looks great!! It's ready to merge as far as I'm concerned. Thanks again for taking the time.

PR #51, opened 2023-03-04
Add support for snapper

Faerbit:
Add support to use the latest snapper snapshot of a source directory.

This addresses [#80](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/80) and [#251](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/251) (though I only tested it with btrfs).

There might still be some usecases which are not adressed by this, but I think it is a good start.

Let me know, if there is anything I should improve upon.

### DIFF
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -109,9 +109,10 @@ def run_create(
         borgmatic.hooks.prepare.PREPARE_HOOK_NAMES,
         location.get('source_directories', []),
     )
-    source_directories = reduce(
-        lambda x, y: x + y if y else x, source_directories_dict.values(), []
-    )
+    # flatten dict with lists as values into one big list
+    source_directories = [
+        directory for directories in source_directories_dict.values() for directory in directories
+    ]
     if source_directories:
         location['source_directories'] = source_directories
 
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -107,13 +107,13 @@ def run_create(
         hooks,
         repository,
         borgmatic.hooks.prepare.PREPARE_HOOK_NAMES,
-        location.get("source_directories", []),
+        location.get('source_directories', []),
     )
     source_directories = reduce(
         lambda x, y: x + y if y else x, source_directories_dict.values(), []
     )
     if source_directories:
-        location["source_directories"] = source_directories
+        location['source_directories'] = source_directories
 
     json_output = borgmatic.borg.create.create_archive(
         global_arguments.dry_run,
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+from functools import reduce
 
 try:
     import importlib_metadata
@@ -13,6 +14,7 @@ import borgmatic.config.validate
 import borgmatic.hooks.command
 import borgmatic.hooks.dispatch
 import borgmatic.hooks.dump
+import borgmatic.hooks.prepare
 
 logger = logging.getLogger(__name__)
 
@@ -99,6 +101,19 @@ def run_create(
         location, global_arguments.used_config_paths, global_arguments.dry_run
     )
     stream_processes = [process for processes in active_dumps.values() for process in processes]
+
+    source_directories_dict = borgmatic.hooks.dispatch.call_hooks_even_if_unconfigured(
+        'prepare_source_directories',
+        hooks,
+        repository,
+        borgmatic.hooks.prepare.PREPARE_HOOK_NAMES,
+        location.get("source_directories", []),
+    )
+    source_directories = reduce(
+        lambda x, y: x + y if y else x, source_directories_dict.values(), []
+    )
+    if source_directories:
+        location["source_directories"] = source_directories
 
     json_output = borgmatic.borg.create.create_archive(
         global_arguments.dry_run,
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -1416,14 +1416,14 @@ properties:
                         description: | 
                             List of source_directories to include or "all" to
                             include all source_directories a snapper config
-                            exists for
+                            exists for.
                     exclude:
                         type: array
                         description: | 
                             List of source_directories to exclude. If a path is
                             listed in include and exclude, exclude will take
-                            precedence
+                            precedence.
                 description: |
                     Configuration for integration with snapper. All included
                     directories will use their latest snapper snapshot for
-                    backup instead of the configured source directory
+                    backup instead of the configured source directory.
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -1402,3 +1402,28 @@ properties:
                     Umask used when executing hooks. Defaults to the umask that
                     borgmatic is run with.
                 example: 0077
+            snapper:
+                type: object
+                required: ['include']
+                additionalProperties: false
+                properties:
+                    include:
+                        anyOf:
+                            - type: array
+                            - type: string
+                              enum:
+                                - all
+                        description: | 
+                            List of source_directories to include or "all" to
+                            include all source_directories a snapper config
+                            exists for
+                    exclude:
+                        type: array
+                        description: | 
+                            List of source_directories to exclude. If a path is
+                            listed in include and exclude, exclude will take
+                            precedence
+                description: |
+                    Configuration for integration with snapper. All included
+                    directories will use their latest snapper snapshot for
+                    backup instead of the configured source directory
--- a/borgmatic/hooks/snapper.py
+++ b/borgmatic/hooks/snapper.py
@@ -69,6 +69,17 @@ def prepare_source_directories(hook_config, _log_prefix, src_dirs):
 
 
 def fix_extracted_dirs(hook_config, log_prefix, src_dirs, destination_path):
+    '''
+    Renames extracted configured source_directories from snapper snapshot path back to their original pre-snapshot path
+    Example: /some/path/.snapshots/3/snapshot to /some/path
+    '''
+    if not src_dirs:
+        logger.warning(
+            f'{log_prefix}: No source_directories configured. Unable to rename snapshot directories. '
+            f'Please restore source_directories config to be same, when creating the archive for best '
+            f'results'
+        )
+        return
     src_dirs = set(map(Path, src_dirs))
     destination_path = Path(destination_path) if destination_path else Path(os.getcwd())
     if hook_config["include"] == "all":
@@ -85,22 +96,25 @@ def fix_extracted_dirs(hook_config, log_prefix, src_dirs, destination_path):
         if snapper_dir.is_absolute():
             snapper_dir = Path(str(snapper_dir)[1:])
         dest_snapper_dir = destination_path / snapper_dir
-        if len(os.listdir(dest_snapper_dir)) != 1:
-            continue
         snap_dir = dest_snapper_dir / ".snapshots"
-        if not snap_dir.is_dir() or len(os.listdir(snap_dir)) != 1:
+        if not snap_dir.is_dir():
             continue
         snap_number_dir = next(snap_dir.iterdir())
         if not snap_number_dir.is_dir() or not snap_number_dir.name.isdigit():
             continue
-
         final_snap_dir = snap_number_dir / "snapshot"
         if not final_snap_dir.is_dir():
             continue
 
-        logger.info(f"{log_prefix}: assuming full system restore: renaming "
-                    f"{final_snap_dir} -> {destination_path / snapper_dir}")
+        if len(os.listdir(dest_snapper_dir)) != 1:
+            logger.warning(
+                f'{log_prefix}: Refusing to overwrite non-empty directory "{dest_snapper_dir}". '
+                f'Run "mv {final_snap_dir} {dest_snapper_dir}" if you want to force this'
+            )
+            continue
+
+        logger.info(f"{log_prefix}: renaming {final_snap_dir} -> {dest_snapper_dir}")
         tmp_snapper_dir = f"{snapper_dir}_"
         final_snap_dir.rename(destination_path / tmp_snapper_dir)
         shutil.rmtree(destination_path / snapper_dir)
-        (destination_path / tmp_snapper_dir).rename(destination_path / snapper_dir)
+        (destination_path / tmp_snapper_dir).rename(dest_snapper_dir)
--- a/borgmatic/hooks/snapper.py
+++ b/borgmatic/hooks/snapper.py
@@ -1,5 +1,7 @@
 import json
 import logging
+import os
+import shutil
 from copy import copy
 from functools import cache
 from pathlib import Path
@@ -64,3 +66,41 @@ def prepare_source_directories(hook_config, _log_prefix, src_dirs):
         processed_dirs.add(snapper_dir)
         altered_dirs.add(new_src_dir)
     return list(map(str, altered_dirs | (src_dirs - processed_dirs)))
+
+
+def fix_extracted_dirs(hook_config, log_prefix, src_dirs, destination_path):
+    src_dirs = set(map(Path, src_dirs))
+    destination_path = Path(destination_path) if destination_path else Path(os.getcwd())
+    if hook_config["include"] == "all":
+        snapper_dirs = copy(src_dirs)
+    else:
+        include_dirs = set(map(Path, hook_config["include"]))
+        snapper_dirs = src_dirs & include_dirs
+    if hook_config.get("exclude"):
+        exclude_dirs = set(map(Path, hook_config["exclude"]))
+        snapper_dirs -= exclude_dirs
+
+    for snapper_dir in snapper_dirs:
+        # remove leading slash to allow joining with other paths
+        if snapper_dir.is_absolute():
+            snapper_dir = Path(str(snapper_dir)[1:])
+        dest_snapper_dir = destination_path / snapper_dir
+        if len(os.listdir(dest_snapper_dir)) != 1:
+            continue
+        snap_dir = dest_snapper_dir / ".snapshots"
+        if not snap_dir.is_dir() or len(os.listdir(snap_dir)) != 1:
+            continue
+        snap_number_dir = next(snap_dir.iterdir())
+        if not snap_number_dir.is_dir() or not snap_number_dir.name.isdigit():
+            continue
+
+        final_snap_dir = snap_number_dir / "snapshot"
+        if not final_snap_dir.is_dir():
+            continue
+
+        logger.info(f"{log_prefix}: assuming full system restore: renaming "
+                    f"{final_snap_dir} -> {destination_path / snapper_dir}")
+        tmp_snapper_dir = f"{snapper_dir}_"
+        final_snap_dir.rename(destination_path / tmp_snapper_dir)
+        shutil.rmtree(destination_path / snapper_dir)
+        (destination_path / tmp_snapper_dir).rename(destination_path / snapper_dir)
--- a/borgmatic/hooks/snapper.py
+++ b/borgmatic/hooks/snapper.py
@@ -1,0 +1,66 @@
+import json
+import logging
+from copy import copy
+from functools import cache
+from pathlib import Path
+
+from borgmatic.execute import execute_command_and_capture_output
+
+logger = logging.getLogger(__name__)
+
+
+def _call_snapper(*args) -> dict:
+    return json.loads(execute_command_and_capture_output(["snapper", "--jsonout", *args]))
+
+
+@cache
+def _available_configs() -> dict[Path, str]:
+    configs = _call_snapper("list-configs")["configs"]
+    # using Path as a key makes it possible to ignore the trailing slash problem
+    # i.e. Path("/test") == Path("/test/")
+    return {Path(c["subvolume"]): c["config"] for c in configs}
+
+
+def prepare_source_directories(hook_config, log_prefix, src_dirs):
+    '''
+    Alter source directory list to use the latest snapper snapshot for each configured path
+    '''
+    if hook_config == {}:
+        return src_dirs
+    src_dirs = set(map(Path, src_dirs))
+    if hook_config["include"] == "all":
+        snapper_dirs = copy(src_dirs)
+        fail = False
+    else:
+        include_dirs = set(map(Path, hook_config["include"]))
+        snapper_dirs = src_dirs & include_dirs
+        fail = True
+    if hook_config.get("exclude"):
+        exclude_dirs = set(map(Path, hook_config["exclude"]))
+        snapper_dirs -= exclude_dirs
+
+    processed_dirs = set()
+    altered_dirs = set()
+    for snapper_dir in snapper_dirs:
+        msg = f'Source directory "{snapper_dir}" was configured to use its latest snapper snapshot for backup, '
+        if snapper_dir not in _available_configs():
+            msg += "but a corresponding snapper config could not be found"
+            if fail:
+                raise ValueError(msg)
+            else:
+                logger.warning(msg)
+                continue
+        config = _available_configs()[snapper_dir]
+        available_snapshots = _call_snapper("-c", config, "list", "--disable-used-space")
+        if not available_snapshots:
+            msg += "but there aren't any snapshots available"
+            if fail:
+                raise ValueError(msg)
+            else:
+                logger.warning(msg)
+                continue
+        latest_snapshot_number = str(available_snapshots[config][-1]["number"])
+        new_src_dir = snapper_dir / ".snapshots" / latest_snapshot_number / "snapshot"
+        processed_dirs.add(snapper_dir)
+        altered_dirs.add(new_src_dir)
+    return list(map(str, altered_dirs | (src_dirs - processed_dirs)))
--- a/borgmatic/hooks/snapper.py
+++ b/borgmatic/hooks/snapper.py
@@ -21,7 +21,7 @@ def _available_configs() -> dict[Path, str]:
     return {Path(c["subvolume"]): c["config"] for c in configs}
 
 
-def prepare_source_directories(hook_config, log_prefix, src_dirs):
+def prepare_source_directories(hook_config, _log_prefix, src_dirs):
     '''
     Alter source directory list to use the latest snapper snapshot for each configured path
     '''
@@ -52,15 +52,15 @@ def prepare_source_directories(hook_config, log_prefix, src_dirs):
                 continue
         config = _available_configs()[snapper_dir]
         available_snapshots = _call_snapper("-c", config, "list", "--disable-used-space")
-        if not available_snapshots:
-            msg += "but there aren't any snapshots available"
-            if fail:
-                raise ValueError(msg)
-            else:
-                logger.warning(msg)
-                continue
         latest_snapshot_number = str(available_snapshots[config][-1]["number"])
         new_src_dir = snapper_dir / ".snapshots" / latest_snapshot_number / "snapshot"
+        if not new_src_dir.exists():
+            msg = (
+                f"Detected snapshot number {latest_snapshot_number} to be the latest for "
+                f"source directory {snapper_dir}, but the deduced directory ({new_src_dir}) is not present. "
+                f"Likely causes are .snapshots not being mounted properly or no snapshots have been taken yet"
+            )
+            raise ValueError(msg)
         processed_dirs.add(snapper_dir)
         altered_dirs.add(new_src_dir)
     return list(map(str, altered_dirs | (src_dirs - processed_dirs)))
--- a/borgmatic/hooks/snapper.py
+++ b/borgmatic/hooks/snapper.py
@@ -3,7 +3,7 @@ import logging
 import os
 import shutil
 from copy import copy
-from functools import cache
+from functools import lru_cache
 from pathlib import Path
 
 from borgmatic.execute import execute_command_and_capture_output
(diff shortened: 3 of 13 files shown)


Faerbit:
Offtopic: I tried creating an account on projects.torsion.org to contribute there, but the registration email doesn't seem to come through.

witten:
Ooh this is exciting! I'll have a look at the PR when I get a chance, but I wanted to let you know that I've manually approved your projects.torsion.org account so you can use that in the future if you like. (I recently moved to a new email server, so some mail is getting delayed.)

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,64 @@
...
+        if snapper_dir not in _available_configs():
+            msg += "but a corresponding snapper config could not be found"
+            if fail:
+                raise ValueError(msg)
+            else:
+                logger.warning(msg)
+                continue
+        config = _available_configs()[snapper_dir]
+        available_snapshots = _call_snapper("-c", config, "list", "--disable-used-space")
+        latest_snapshot_number = str(available_snapshots[config][-1]["number"])
```
witten, line 55: Is any error handling needed here?

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,64 @@
...
+        new_src_dir = snapper_dir / ".snapshots" / latest_snapshot_number / "snapshot"
+        if not new_src_dir.exists():
+            msg = f"Detected snapshot number {latest_snapshot_number} to be the latest for " \
+                  f"source directory {snapper_dir}, but the deduced directory ({new_src_dir}) is not present. " \
+                  f"Likely causes are .snapshots not being mounted properly or no snapshots have been taken yet"
```
witten, line 60: So is borgmatic responsible for triggering the snapshot to be taken? Or does that happen elsewhere?

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,64 @@
...
+            raise ValueError(msg)
+        processed_dirs.add(snapper_dir)
+        altered_dirs.add(new_src_dir)
+    return list(map(str, altered_dirs | (src_dirs - processed_dirs)))
```
witten, line 64: Similarly, I recommend using set methods instead of operators here for clarity.

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,64 @@
...
+    if hook_config.get("exclude"):
+        exclude_dirs = set(map(Path, hook_config["exclude"]))
+        snapper_dirs -= exclude_dirs
+
+    processed_dirs = set()
+    altered_dirs = set()
+    for snapper_dir in snapper_dirs:
+        msg = f'Source directory "{snapper_dir}" was configured to use its latest snapper snapshot for backup, '
+        if snapper_dir not in _available_configs():
+            msg += "but a corresponding snapper config could not be found"
```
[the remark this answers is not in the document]
witten: And what if only a parent or grandparent directory is configured for snapshotting at the Snapper level? For instance, let's say an included source directory is `/foo/bar/baz` but only `/` is configured for snapshotting within Snapper. Will that not work?

witten (changes requested):
Partial review: Everything but tests.. I wanted to get something to you sooner rather than later.

Overall, this approach looks great. You seem to have done a nicely surgical bait-and-switch on the borgmatic source directories to get the snapshotted variants backed up as requested. If it seems like I left a lot of comments on the PR, it's because I did. :smile: But most of them fall into one of two categories:

1. Minor code style stuff, hopefully super easy to either fix or ignore as appropriate.
2. Me just trying to understand the general scope of the solution. I certainly don't expect all of it to be encoded here in docstrings or schema comments. But if I eventually might be on the hook for documenting and/or supporting this feature, I better understand how it's intended to work!

So I'm going to outline my understanding, and then you can correct me where I go off the rails.

A user starts by needing to backup a snapshot of a source directory rather than a bare source directory itself. Maybe they want a consistent snapshot so things aren't changing during the backup. Whatever the reason, they configure the source directory in the `includes` for the new borgmatic `snapper:` hook, and also configure that directory in some external snapper configuration. (`snapper -c ... create-config ...`?)

Then whenever borgmatic runs a `create` action, the hook swaps out the requested source directory for a `.snapshot` sub-directory containing the snapshot. And I guess something external to borgmatic takes the snapshot (already took the snapshot)? I'm kind of hazy on how/where that happens.

Does that sound approximately right? Any help you can provide with my understanding would be great. And I really appreciate you taking the time to put together this PR and submit it. I think this is going to be a much-needed addition to borgmatic.

Faerbit:
Maybe read the top half of [this](https://wiki.archlinux.org/title/Snapper) to understand how one might want to use snapper. But I'll try to give some basic explanation:

Basically, there are two and a half ways to use snapper:
* Timeline snapshots
* Integration with system management (mostly package management, but I think Suse has more advanced stuff)
* Your own custom usage calling the snapper `create` action directly with some script/cron job/systemd timer ...

So for timeline snapshots, there is a systemd timer running taking a snapshot every hour (upstream default).
For package management (e.g. apt, dnf, pacman) integration, there is a snapshot taken before pacman is run, and after that (to revert an upgrade having a snapshot from right before the upgrade would suffice, but I think it makes for cleaner diffs to have the upgrades nicely encapsulated).

So to answer your question: Yes, snapper is responsible for taking the snapshots beforehand and does so regardless of borg/borgmatic coming around later to pick them up. Otherwise your understanding is pretty good I think.

Depending on what the user wants exactly, just taking the latest snapshot, might not be enough (maybe they prefer the snapshot from before the system upgrade). But since I have a hard time imagining what they might want (and personally I am content with just having the latest snapshot), it might make sense to start like this, and then wait for user feedback and take it from there.

I'll try to address all of your comments, might take me a bit :)

How is review etiquette around here? If I address a comment, should I resolve it if I feel I addressed it completely, or just wait for you to close it?

Thanks for being so welcoming :)

borgmatic/actions/create.py
```diff
@@ -55,6 +57,19 @@ def run_create(
     )
     stream_processes = [process for processes in active_dumps.values() for process in processes]
 
+    source_directories_dict = borgmatic.hooks.dispatch.call_hooks_even_if_unconfigured(
```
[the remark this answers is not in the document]
Faerbit: I had it like this in my first iteration. The problem is, that one needs to figure out what source directories should be kept unaltered (if only a subset has snapshots). Personally, I would put that code into the hook, which is why it needs to called unconditionally.

But I am open to an alternate code structure :)

borgmatic/config/schema.yaml
```diff
@@ -1265,3 +1265,28 @@ properties:
                     Umask used when executing hooks. Defaults to the umask that
                     borgmatic is run with.
                 example: 0077
+            snapper:
+                type: object
+                required: ['include']
+                additionalProperties: false
+                properties:
+                    include:
```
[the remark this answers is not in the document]
Faerbit: I think a good example would be that one has a btrfs filesystem and an ext4 filesystem and wants to backup both of them. Mainly I implemented it like this because it seemed way too inflexible to me to just have this as an all or nothing option.

borgmatic/config/schema.yaml
```diff
@@ -1265,3 +1265,28 @@ properties:
...
+                    exclude:
+                        type: array
+                        description: | 
+                            List of source_directories to exclude. If a path is
+                            listed in include and exclude, exclude will take
+                            precedence
+                description: |
+                    Configuration for integration with snapper. All included
+                    directories will use their latest snapper snapshot for
+                    backup instead of the configured source directory
```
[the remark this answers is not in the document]
Faerbit: Hmm, so this is where it gets ugly apparently :D
>Do you have thoughts on whether users will care that their snapshotted files will effectively be in a different location in the backup archive?

Yeah, this seems pretty annoying. One way to address this might be to call borg like this: `subprocess.run["borg", "create", repo, ".", cwd=source_directory` for each source directory. But I suppose this is completely different from how borgmatic does things currently.

With regards to the database dumps: Since the snapshots are taken before the dumps are created, I'm not sure how that would integrate with borgmatic at all. Excluding `~/.borgmatic` from snapshotting can not be done within borgmatic, but needs to be done in snapper config/btrfs subvolumes/mount hierarchy. This could be helped through documentation though I suppose.

But additionally `~/.borgmatic` would need to be readded to the source_directories such that the dumps will be included.

I need to think on this a bit but feel free to chime in with insights of your own if you have any.

borgmatic/config/schema.yaml
```diff
@@ -1265,3 +1265,28 @@ properties:
                     Umask used when executing hooks. Defaults to the umask that
                     borgmatic is run with.
                 example: 0077
+            snapper:
```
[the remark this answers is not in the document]
Faerbit: Snapper is the *only* tool that does something like this (AFAIK), but not *the* tool (IMO). ZFS support is [apparently](https://github.com/openSUSE/snapper/issues/145) somewhat on the way but seems pretty stalled.
So it seems possible for something like `zfs_snapshots` to make sense in the future.

I disagree with your conclusion though: I would call it `snapper` and `zfs_snapshots` since I don't find these suffixes particularly helpful. I would also point to `cronitor`, `healthchecks`, `pagerduty`, ... not having a common suffix (e.g. `_monitoring`).

But ultimately this is your thing, and I am fine with whatever you want this to be named.

borgmatic/actions/create.py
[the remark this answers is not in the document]
witten: I might be misunderstanding something: If `snapper:` is completely omitted from configuration, then that should effectively disable the feature, right? Such that no mutation of the source directories needs to happen. But then as soon as `snapper:` is added to configuration, the source directory munging needs to kick in. Which is what would happen with the plain `call_hooks()`...?

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: Makes sense.. Thanks for explaining. Do you have any thoughts on how this could/should interact with the built-in Borg/borgmatic excludes/patterns functionality? For instance, since all of this processing on `source_directories` happens before Borg is called, you could end up converting some excluded source directories into snapshot paths—which would then potentially no longer get excluded. Similarly, configured `patterns` could effectively include additional directories that can't use the snapper integration because they're not in `source_directories`.

Maybe these are super edge cases? Perhaps just known limitations to document?

witten:
Your explanation makes sense to me.. Much appreciated! (I should've known it'd be explained on the ArchWiki.) Simply using the latest snapshot makes sense to me. In fact, why not call `snapper create` unconditionally in borgmatic to: 1. Ensure there's actually a most recent backup, and 2. Make it as close to the backup time as possible. Is the idea that the timeline snapshots are "good enough?"

> How is review etiquette around here? If I address a comment, should I resolve it if I feel I addressed it completely, or just wait for you to close it?

I subscribe to the "we're all adults" review etiquette: Resolve it if you feel like it's resolved; leave it open otherwise. I can always reopen if needed. Thanks for asking. :smiley:

witten:
Here's an idea that would be great for you to consider, but feel free to disregard it if you don't think the benefits are worth it: In some of the other tickets about borgmatic snapshotting features, there have been conceptual discussions of using [bind mounts](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/261#issuecomment-2587) to [present to Borg](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/80#issuecomment-2072) the illusion that snapshots have the exact same path as the source directories they're snapshots of. Put another way, instead of having to rewrite `/source/dir` to `/source/dir/.snapshots/5/snapshot` or whatever, you would bind mount `/source/dir/.snapshots/5/snapshot` *onto* `/source/dir` just for the borgmatic/Borg process so that it got backed up as if the path hadn't changed.

This would have the benefit that all the existing Borg excludes/patterns worked as expected, you could maybe support *both* `source_directories` and `patterns` depending on how you did it, paths within Borg archives would be where users expected when they went to do restores, etc etc. Downsides: Potentially more complicated to implement, not the approach you already coded up, probably root-only.

Anyway, let me know your thoughts.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: > Yeah, this seems pretty annoying.

I left an idea in another comment about bind mounts that might help address this.

> One way to address this might be to call borg like this: subprocess.run(["borg", "create", repo, "."], cwd=source_directory) for each source directory. But I suppose this is completely different from how borgmatic does things currently.

Interesting! Yeah, I think that would also result in the backed up path showing up differently in the archive. Because today IIRC the full `/source/directory/` is a prefix on every path in the archive.

> I need to think on [the database dump question] a bit but feel free to chime in with insights of your own if you have any.

Bind mount? :smiley: Then it doesn't matter if `~/.borgmatic` is snapshotted because the path won't change. (I admit I have a :hammer: and everything looks like a nail.)

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: Nah, I don't feel strongly here. IIRC for the database hooks, the `_databases` suffix was less to serve as a crappy namespace and more because of what each one is listing is.. databases! But in the case of your hook, you're not listing snapshots, per se. So :shrug:.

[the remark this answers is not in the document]
Hofer-Julian: > Snapper is the only tool that does something like this

That's probably correct when it comes to root snapshots.
For general snapshots, [btrbk](https://github.com/digint/btrbk) comes to mind

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+        (
+            {"include": "all", "exclude": ["/baz"]},
+            ["/foo", "/bar", "/baz"],
+            ["/foo/.snapshots/17/snapshot", "/bar/.snapshots/1337/snapshot", "/baz"],
+        ),
+        (
+            {"include": "all", "exclude": ["/baz"]},
+            ["/foo", "/bar", "/baz"],
+            ["/foo/.snapshots/17/snapshot", "/bar/.snapshots/1337/snapshot", "/baz"],
+        ),
```
witten, line 97: Unless I'm missing something, this looks like duplicated test parameters. (Same as the last one.)

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+            ["/foo", "/bar", "/baz"],
+        ),
+        (
+            {"include": "all"},
+            ["/foo", "/bar", "/baz", "/quaz"],
+            [
+                "/foo/.snapshots/17/snapshot",
+                "/bar/.snapshots/1337/snapshot",
+                "/baz/.snapshots/13337/snapshot",
+                "/quaz",
```
witten, line 120: I always understood `quux` to be next after `baz`. Different schools of thought I guess. :smile:

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+                        "number": 13337,
+                    },
+                ]
+            }
+        )
+    )
+    flexmock(module.Path).should_receive("exists").and_return(True)
+
+    src_dirs_after = module.prepare_source_directories(config, log_prefix, src_dirs_in)
+    assert set(src_dirs_after) == set(expected)
```
witten, line 192: Also a nice-looking test! Seems to cover the bulk of the possible inputs/states.

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+                    },
+                ],
+            }
+        )
+    )
+
+    src_dirs_before = ["/foo"]
+    with pytest.raises(ValueError) as e_info:
+        module.prepare_source_directories({"include": ["/foo"]}, log_prefix, src_dirs_before)
+    assert "could not be found" in str(e_info)
```
witten, line 216: This assertion might need indenting under the `with` so `e_info` is always set? For instance, does it get set if there's no exception? I don't think there's risk of exiting the function cleanly if there's no `ValueError` exception because of that `pytest.raises()`.

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+            }
+        )
+    )
+    flexmock(module.Path).should_receive("exists").once().and_return(False)
+
+    src_dirs_before = ["/foo"]
+    with pytest.raises(ValueError) as e_info:
+        module.prepare_source_directories({"include": "all"}, log_prefix, src_dirs_before)
+
+    assert re.match(r".*deduced directory .* is not present.*", str(e_info))
```
witten, line 265: Same comment here about indenting.

witten (commented):
Tests look good! Let me know what direction you'd like to go with this feature, and I'd be happy to discuss and/or review subsequent PR updates.

Faerbit:
I thought about your bind mount approach. What I dislike about it, is that it doesn't allow for restoring a backup, in a running system:

I.e. you want to restore a snapshot, which you do not have locally anymore, only in the backup. Either to just have a look at it or to restore to it, while keeping the other local snapshots.

That is why I implemented a middle ground: When a full system restore is detected (the snapshot directory is the only one present) the snapshot directory is renamed to be the source directory. Having the same borgmatic config is necessary for this to work. 

What do you think about this approach?

If you like it, I will add a test for it and address the rest of your comments.

witten:
I'm totally fine with your proposed approach, but I have a few comments:

> I.e. you want to restore a snapshot, which you do not have locally anymore, only in the backup. Either to just have a look at it or to restore to it, while keeping the other local snapshots.

Ah okay, so it sounds like you're not thinking of snapshots as this ephemeral thing strictly for purposes of atomic backups, but rather as first-class entities that you might want backed up and restored themselves. Your approach would certainly support that!

> That is why I implemented a middle ground: When a full system restore is detected (the snapshot directory is the only one present) the snapshot directory is renamed to be the source directory. Having the same borgmatic config is necessary for this to work.

That all makes sense, but I wouldn't necessarily call that a "full system restore." You might just be restoring a single accidentally deleted directory that happens to have its own snapshots configured, and the rest of the system is left untouched.

Unrelated, an idea: Instead of or in addition to auto-detecting the "rename the snapshot onto the original after extraction" use case, you _could_ make that behavior explicit. Like there's a "borgmatic extract --rename-snapshots" flag or whatever that triggers that behavior explicitly. Up to you though.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
shalearkane: The bind mount idea is the best. It will also help in data deduplication and faster archive creation. However, will it be prudent to also store the absolute directory path in the comment (or in the metadata) of the archive? Because then we will be able to restore to the exact location.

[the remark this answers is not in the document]
witten: As much as I like the idea of bind mounts, I'm not sure that it would impact deduplication. My understanding is that Borg deduplicates at the [chunk level](https://borgbackup.readthedocs.io/en/stable/#main-features) regardless of filename.

witten:
No worries. I appreciate the update. The only risk with a limbo state is that this branch will drift further from master. If you're okay with that, I am too!

witten:
Hey @Faerbit, just checking in on this one. I imagine you're still occupied with exciting things, which is great. But do you intend to get back to this PR at some point? Either way, please let me know! Thanks.

borgmatic/hooks/snapper.py
Faerbit: Usually that happens via a cron job / systemd timer or triggered by some other event (e.g. package manager doing stuff).
Leaving this separate gives the user flexibility to configure this however they want to.

One could trigger snapper snapshots from borgmatic, but that only works as a root user, even though borgmatic could run as a user. So that complicates things

Additionally I don't see any usecase, which can't be implemented by correctly timed invocations of snapper / borgmatic.

borgmatic/hooks/snapper.py
Faerbit: done

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+from borgmatic.hooks import snapper as module
+
+
+@pytest.fixture(scope="function", autouse=True)
+def setup():
+    module._available_configs.cache_clear()
+
+
+@pytest.mark.parametrize("src_dirs_before", ([], ["/foo"], ["/foo", "/bar", "/baz"]))
+def test_unaltered_if_not_configured(src_dirs_before):
```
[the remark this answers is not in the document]
Faerbit: done

tests/unit/hooks/test_snapper.py
Faerbit: you're right. Sharp eyes :)

tests/unit/hooks/test_snapper.py
Faerbit: I have always written my tests like this, and I think due to how python handles variable scoping, this is functionally equivalent. However your approach seems to be more clean, so I changed it.

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+    log_prefix = ''
+    src_dirs_after = module.prepare_source_directories({}, log_prefix, src_dirs_before)
```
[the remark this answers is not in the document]
Faerbit: Your assumption is correct. Should I change it?

tests/unit/hooks/test_snapper.py
```diff
@@ -0,0 +1,265 @@
...
+            }
+        )
+    )
+    flexmock(module.Path).should_receive("exists").once().and_return(True)
+
+    src_dirs_before = ["/foo"]
+    src_dirs_after = module.prepare_source_directories(
+        {"include": "all"}, log_prefix, src_dirs_before
+    )
+    assert src_dirs_after == ["/foo/.snapshots/17/snapshot"]
```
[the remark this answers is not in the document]
Faerbit: > I do wonder though why this test isn't just shoved into the parametrized test function following this one.. Seems pretty similar.

I like having a simple as possible test, such that one can understand how the function works without any ugly edgecases.

> just to keep the unit under test as small and independent as possible

I'm not as puristic about this, but I can see where this is coming from. Should I change it?

borgmatic/hooks/snapper.py
Faerbit: I wasn't able to get snapper into a state where no snapshot was available (it already starts out with one and even refuses to delete that). So I would say we are reasonably safe here.

Faerbit (commented):
Sorry, for the long pause I addressed most (all?) of your concerns and implemented the optional argument for renaming extracted dirs, because I liked it better than the automagic, which I had before.

I need some guidance on the testing for that though:
The function `fix_extracted_dirs` has many filesystem calls, which would be quite cumbersome (and also quite unhelpful imo) to mock out.
Usually I would just copy some actual files to some temp dir and do my testing there.
But I get the feeling that's not how are things are done here.

So how would you go about testing this function?

witten:
Thank you for getting back to this! Just wanted to let you know I'll look into your responses and provide my thoughts as soon as I get a chance.

witten:
> I need some guidance on the testing for that though:
> The function fix_extracted_dirs has many filesystem calls, which would be quite cumbersome (and also quite unhelpful imo) to mock out.
> Usually I would just copy some actual files to some temp dir and do my testing there.
> But I get the feeling that's not how are things are done here.
> 
> So how would you go about testing this function?

I think your strategy with a temporary directory makes sense for this particular function. The way I'd do that in borgmatic is with an "end-to-end" test. See `tests/end-to-end/` for some examples and also the docs here for how to run these tests: https://torsion.org/borgmatic/docs/how-to/develop-on-borgmatic/#end-to-end-tests

Let me know if you have any questions on any of that. There _does_ seem to be a fair amount of logic within `fix_extracted_dirs()` though, so if it were me I'd probably at least try to do some mock-heavy unit tests to cover most of the logic. And then I'd also do one or two end-to-end tests with a temp directory (without having to test 100% of the logic in them).

witten:
A couple more thoughts on the `fix_extracted_dirs()` tests topic. If you write an end-to-end test for this, it should probably trigger borgmatic as a whole (being an end-to-end test and all) ... first adjusting the configuration it consumes such that your desired code path gets hit and tested. You can see why this approach is not so great for testing all the permutations of a function's logic though. It'd be like trying to verify the ordering in a deck of a cards while wearing boxing gloves.

But then, going back to unit tests for testing the logic, you're faced with the problem of heavy mocking again. There is an alternate though: Refactoring the `fix_extracted_dirs()` function a bit (well, maybe more than a bit) to separate out the logic from the side-effecting code. Maybe even splitting those into two different functions, respectively. So maybe (just an example), you could have one function that takes as input a recursive file/directory structure as previously read from disk, does *logic* (without hitting disk), and then returns a list of renames that need doing (but without doing them). That function would be purely functional and presumably very easy to unit test. And then you could have your separate disk-hitting code be relatively logic-free and easier to test with end-to-end tests or whatever.

Faerbit:
> There is an alternate though: Refactoring the fix_extracted_dirs() function a bit (well, maybe more than a bit) to separate out the logic from the side-effecting code

This was my first thought as well, but I didn't see any obvious way to do this.

>you could have one function that takes as input a recursive file/directory structure as previously read from disk,

This would work, but this would mean walking down the *entire* directory tree, when we are only interested in some directories at the top. I would imagine that this wouldn't perform well for larges filesystems. Again: I get where you're coming from in principle, but I don't see any good way to apply this logic here.

Anyway: After a bit of struggling I managed to put together an end-to-end test.
It works and I think it demonstrates what is happening quite good. There are several caveats though:
* snapper is only available in the edge repository for Alpine
* To have some snapshots, I needed to have a filesystem, which I accomplished using a loop device. This needs a privileged container running with root privileges to work though, so this not quite optimal

One approach to address these problems could be to separate them from the rest of the end-to-end test. E.g. have this one running under Debian for better support of snapper. 

I'm not sure what to do about the root thing, I would think that most people do not like having to run a test suite as root.

Let me know your thoughts :)

witten:
> >    you could have one function that takes as input a recursive file/directory structure as previously read from disk,
>
> This would work, but this would mean walking down the entire directory tree, when we are only interested in some directories at the top. I would imagine that this wouldn't perform well for larges filesystems. Again: I get where you're coming from in principle, but I don't see any good way to apply this logic here.

Fair enough, although I can imagine a version that only descends into the collected `snapper_dirs` and their immediate children just like the existing code is doing. If I'm reading `fix_extracted_dirs()` correctly, the production of `snapper_dirs` itself doesn't hit the filesystem, and the subsequent loop over them doesn't descend past the immediate children (+ those of the destination `.snapshots` paths).

> Anyway: After a bit of struggling I managed to put together an end-to-end test.
It works and I think it demonstrates what is happening quite good. There are several caveats though:

>     snapper is only available in the edge repository for Alpine

I'm fine switching to edge (for now) as long as that doesn't break the other end-to-end tests.

>     To have some snapshots, I needed to have a filesystem, which I accomplished using a loop device. This needs a privileged container running with root privileges to work though, so this not quite optimal

This I'm less wild about. :smiley:

> One approach to address these problems could be to separate them from the rest of the end-to-end test. E.g. have this one running under Debian for better support of snapper.
>
> I'm not sure what to do about the root thing, I would think that most people do not like having to run a test suite as root.

Yeah, I think these are the options as I see them:

 * Run in a container as `privileged` as you do here, although that poses some challenges. For instance, it won't work in rootless Docker, rootless Podman, or even possibly in the existing CI system. This could maybe be documented around though.
 * Run this test directly on the host system. This might work in CI as-is? (CI does not currently use container-in-container.)
 * Use real virtualization like qemu. That would allow rootless testing—at the cost of yet more test infrastructure for devs to get working. And no idea how this would work in CI.

Of those options, I think your current privileged container approach (+ docs changes) would probably be best. And I don't think it's necessary to break out the filesystem tests into separate end-to-end tests that run privileged as long as they can all be run together with Compose; probably anyone bothering to run end-to-end tests will be okay with running them all together (as privileged).

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+        snapper_dirs = copy(source_directories)
+        include_all = True
+    else:
+        include_dirs = set(map(Path, hook_config['include']))
+        snapper_dirs = source_directories.intersection(include_dirs)
+        include_all = False
+
+    if hook_config.get('exclude'):
+        exclude_dirs = set(map(Path, hook_config['exclude']))
+        snapper_dirs -= exclude_dirs
```
witten, line 48: It occurs to me that if you lean on a Borg dry-run to produce the list of directories to process, you could even make it apply the includes and excludes for you so you don't have to here. You'd just have to pass them in on Borg's command-line.

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+    '''
+    Renames extracted configured source_directories from snapper snapshot path back to their original pre-snapshot path
+    Example: /some/path/.snapshots/3/snapshot to /some/path
+    '''
+
+    if not source_directories:
+        logger.warning(
+            f'{log_prefix}: No source_directories configured. Unable to rename snapshot directories. '
+            f'Please restore source_directories config to be same, when creating the archive for best '
+            f'results'
```
witten, line 91: I don't quite follow the final sentence here. Perhaps it could be rephrased?

Also, as mentioned above, not having any source directories is unfortunately a valid borgmatic (and Borg) configuration. But maybe the ol' Borg dry-run file list could help this use case too? That way you could use the Borg-computed result of whatever source directories, patterns, excludes, etc. are configured in borgmatic (including maybe these Snapper-specific includes/excludes).

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+        final_snap_dir = snap_number_dir / 'snapshot'
+
+        if not final_snap_dir.is_dir():
+            continue
+
+        if len(os.listdir(dest_snapper_dir)) != 1:
+            logger.warning(
+                f'{log_prefix}: Refusing to overwrite non-empty directory "{dest_snapper_dir}". '
+                f'Run "mv {final_snap_dir} {dest_snapper_dir}" if you want to force this'
+            )
```
witten, line 130: Smart.

The one thing I don't follow is the `!= 1` part. Is that because the directory is empty except for a Snapper dotfile or something? How do you know that in particular is what's in there?

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+        exclude_dirs = set(map(Path, hook_config['exclude']))
+        snapper_dirs -= exclude_dirs
+
+    for snapper_dir in snapper_dirs:
+        # remove leading slash to allow joining with other paths
+
+        if snapper_dir.is_absolute():
+            snapper_dir = Path(str(snapper_dir)[1:])
+        dest_snapper_dir = destination_path / snapper_dir
+        snap_dir = dest_snapper_dir / '.snapshots'
```
witten, line 113: Variable naming between `snapper_dir` and `snap_dir` is a little confusing to me. Reading how they're defined, I can see the difference between them. But especially in a function/loop with so many different directory variables, more distinct variable naming would help a ton.

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+            logger.warning(
+                f'{log_prefix}: Refusing to overwrite non-empty directory "{dest_snapper_dir}". '
+                f'Run "mv {final_snap_dir} {dest_snapper_dir}" if you want to force this'
+            )
+            continue
+
+        logger.info(f'{log_prefix}: renaming {final_snap_dir} -> {dest_snapper_dir}')
+        tmp_snapper_dir = f'{snapper_dir}_'
+        final_snap_dir.rename(destination_path / tmp_snapper_dir)
+        shutil.rmtree(destination_path / snapper_dir)
```
witten, line 136: Isn't this just `dest_snapper_dir`?

borgmatic/hooks/snapper.py
```diff
@@ -0,0 +1,137 @@
...
+        if len(os.listdir(dest_snapper_dir)) != 1:
+            logger.warning(
+                f'{log_prefix}: Refusing to overwrite non-empty directory "{dest_snapper_dir}". '
+                f'Run "mv {final_snap_dir} {dest_snapper_dir}" if you want to force this'
+            )
+            continue
+
+        logger.info(f'{log_prefix}: renaming {final_snap_dir} -> {dest_snapper_dir}')
+        tmp_snapper_dir = f'{snapper_dir}_'
+        final_snap_dir.rename(destination_path / tmp_snapper_dir)
```
witten, line 135: Why the rename to a temporary directory?

witten (commented):
This is a partial review, because I figured getting you _some_ comments was better than continuing to sit on this PR!

Faerbit:
I am really burnt out on this and therefore won't continue this. Your expections, both feature- and code qualitywise far exceed what constitues a fun side-project for my free time. I am not saying it is wrong to have this expectations, I just am not able to meet them. Sorry.

I'll gladly try to test whatever comes out of this (if anything), or provide feedback if you want me to.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
Faerbit: I think your `--dry-run` approach is a horrible idea performance-wise, but the alternative (handling parsing of the patterns, and matching that with available snapper configs) is a horrible idea code complexity-wise.
So I don't know what a good solution would be.

borgmatic/hooks/snapper.py
Faerbit: Python simply can't handle a direct replacement like `mv` can. I don't know why and was confused as well ;)

borgmatic/hooks/snapper.py
Faerbit: There should only be a ` .snapshot` dir (containing the snapshot) present, otherwise something funky happened. I assume this, because this is how it was prepared in the `prepare_source_directories` function.

borgmatic/hooks/snapper.py
Faerbit: If the borgmatic config is not the same, when restoring, as when backing up, I have no idea of how to get any proper results.

I have no clue of how this should be handled when only using patterns.

witten:
I'm sorry to hear that, but I also totally understand! I think big features like this are particularly tough to complete in open source projects for exactly the reasons you mention—you're doing this in your free time for fun and not getting paid. I do appreciate all the work and re-work you've put into this PR, and hopefully some of it can serve as inspiration for subsequent efforts on borgmatic snapshot support.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: Yeah, it's a thorny problem with this approach, and I'm not sure what I'd do there to solve it. The problem is that Borg basically has its own domain-specific language for what directories should be backed up. And if you'd like to _operate_ on that set of directories, you either need to get into the business of interpreting the domain-specific language or lean on Borg to do it for you.

borgmatic/hooks/snapper.py
witten: TIL!

borgmatic/hooks/snapper.py
witten: Got it.. Thanks for explaining.

PR #52, opened 2023-03-04
fix: remove extra dark mode styles

diivi:
To avoid unintended behaviour on devices that have enabled dark mode.


witten:
Cool, thanks for doing this!

PR #55, opened 2023-03-21
fix: rephrase error when running from config

diivi:
to avoid confusion, as the user might think the problem is with their config file.

fixes [#623](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/623)


witten:
Thank you!

PR #56, opened 2023-03-22
fix: no error on database backups without source dirs

diivi:
Your detailed explanation in [#655](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/655#issuecomment-5679) helped.

This solution passes all current tests, so it should be backward compatible, and does not raise an error anymore when source dirs don't exist:
image

### DIFF
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -291,6 +291,7 @@ def collect_special_file_paths(
         capture_stderr=True,
         working_directory=working_directory,
         extra_environment=borg_environment,
+        raise_on_exit_code_one=False,
     )
 
     paths = tuple(
--- a/borgmatic/execute.py
+++ b/borgmatic/execute.py
@@ -213,7 +213,12 @@ def execute_command(
 
 
 def execute_command_and_capture_output(
-    full_command, capture_stderr=False, shell=False, extra_environment=None, working_directory=None,
+    full_command,
+    capture_stderr=False,
+    shell=False,
+    extra_environment=None,
+    working_directory=None,
+    raise_on_exit_code_one=True,
 ):
     '''
     Execute the given command (a sequence of command/argument strings), capturing and returning its
@@ -221,20 +226,29 @@ def execute_command_and_capture_output(
     stdout. If shell is True, execute the command within a shell. If an extra environment dict is
     given, then use it to augment the current environment, and pass the result into the command. If
     a working directory is given, use that as the present working directory when running the command.
+    If raise on exit code one is False, then treat exit code 1 as a warning instead of an error.
 
-    Raise subprocesses.CalledProcessError if an error occurs while running the command.
+    Raise subprocesses.CalledProcessError if an error occurs while running the command, or if the
+    command exits with a non-zero exit code and raise on exit code one is True.
     '''
     log_command(full_command)
     environment = {**os.environ, **extra_environment} if extra_environment else None
     command = ' '.join(full_command) if shell else full_command
 
-    output = subprocess.check_output(
-        command,
-        stderr=subprocess.STDOUT if capture_stderr else None,
-        shell=shell,
-        env=environment,
-        cwd=working_directory,
-    )
+    try:
+        output = subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT if capture_stderr else None,
+            shell=shell,
+            env=environment,
+            cwd=working_directory,
+        )
+        logger.warning('Command output: {}'.format(output))
+    except subprocess.CalledProcessError as e:
+        if raise_on_exit_code_one or e.returncode != 1:
+            raise
+        output = e.output
+        logger.warning('Command output: {}'.format(output))
 
     return output.decode() if output is not None else None
 
--- a/tests/unit/test_execute.py
+++ b/tests/unit/test_execute.py
@@ -239,6 +239,32 @@ def test_execute_command_and_capture_output_with_capture_stderr_returns_stderr()
     assert output == expected_output
 
 
+def test_execute_command_and_capture_output_returns_output_with_raise_on_exit_code_one_false():
+    full_command = ['foo', 'bar']
+    expected_output = '[]'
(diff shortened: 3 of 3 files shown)


borgmatic/execute.py
```diff
@@ -213,28 +213,42 @@ def execute_command(
...
+    try:
+        output = subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT if capture_stderr else None,
+            shell=shell,
+            env=environment,
+            cwd=working_directory,
+        )
+        logger.warning('Command output: {}'.format(output))
+    except subprocess.CalledProcessError as e:
```
witten, line 247: Minor nit: The (unwritten) code style is that borgmatic source doesn't use single-letter variables.

borgmatic/borg/create.py
```diff
@@ -291,6 +291,7 @@ def collect_special_file_paths(
         capture_stderr=True,
         working_directory=working_directory,
         extra_environment=borg_environment,
+        raise_on_exit_code_one=False,
```
witten, line 294: Adding this parameter to `execute_command_and_capture_output()` is a clever way to change the error behavior of just `collect_special_file_paths()` without affecting other code, but the name of the parameter "leaks" implementation details about what's considered an error. Maybe it could be named something more generic like `raise_on_exit_code_warning` or `treat_exit_code_warning_as_error` or `error_on_warning` or whatever.

I also wonder a little whether the parameter is needed at all. Maybe this same type of bug exists in some of the other callers of `execute_command_and_capture_output()`.... Having wondered that aloud, I don't think you need to expand scope to encompass the other callers right now, but it's maybe something to think about or look out for in the future.

borgmatic/execute.py
```diff
@@ -213,28 +213,42 @@ def execute_command(
...
+        if raise_on_exit_code_one or e.returncode != 1:
+            raise
```
witten, line 249: This logic looks correct to me, but it's effectively repeating some of the logic encapsulated in `exit_code_indicates_error()`, which has indeed changed over time as the concept of "what's an error" has evolved with different Borg releases. So it would be great to continue to keep that logic in one place.

To do that, maybe you could call `exit_code_indicates_error()` as needed from `execute_command_and_capture_output()`? Right now `exit_code_indicates_error()` takes a `process` as a parameter, but there's no reason it needs to. It could just as easily take a command sequence (which you have here), and then other callers could pass in `process.args` for that command.

I know this expands scope a little, but hopefully not too much as it's mostly a mechanical change.

witten (changes requested):
This looks like a good start! A few notes on the details are below.

borgmatic/borg/create.py
witten: borgmatic has 100%-ish test coverage, but it's mostly unit tests instead of integration tests. So if some callers in one unit are implicitly relying on Borg warnings being treated as errors in another unit, that won't necessarily be captured in any existing test.

Having said that, I went ahead and manually reviewed all callers of `execute_command_and_capture_output()`, and I didn't immediately see anything that would be problematic. And just thinking this through, if some existing code calls Borg and now doesn't receive an exception when Borg returns exit code 1, that should hopefully not hurt anything and is arguably more correct.

The one case I did test manually was when `borgmatic init` probes for repository existence with `borg rinfo`/`borg info`. And I did confirm that when the repository doesn't exist, Borg helpfully returns exit code 2 instead of 1.

So my recommendation is to proceed with your approach!

borgmatic/execute.py
```diff
@@ -221,20 +220,29 @@ def execute_command_and_capture_output(
     stdout. If shell is True, execute the command within a shell. If an extra environment dict is
     given, then use it to augment the current environment, and pass the result into the command. If
     a working directory is given, use that as the present working directory when running the command.
+    If raise on exit code one is False, then treat exit code 1 as a warning instead of an error.
```
witten, line 223: Looks like this docstring comment is now out of date!

borgmatic/execute.py
```diff
@@ -221,20 +220,29 @@ def execute_command_and_capture_output(
...
 
-    Raise subprocesses.CalledProcessError if an error occurs while running the command.
+    Raise subprocesses.CalledProcessError if an error occurs while running the command, or if the
+    command exits with a non-zero exit code and raise on exit code one is True.
```
witten, line 226: It looks like this comment is also out of date. I think you can almost just leave it as it was: `Raise subprocesses.CalledProcessError if an error occurs while running the command.` That's because an error is non-zero for all commands except for Borg, for which warnings (exit code 1) are treated as such and errors (not exit code 1 or 0) are treated as errors. So in all cases, you're raising for semantic errors.

borgmatic/execute.py
```diff
@@ -221,20 +220,29 @@ def execute_command_and_capture_output(
...
+        output = subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT if capture_stderr else None,
+            shell=shell,
+            env=environment,
+            cwd=working_directory,
+        )
+        logger.warning('Command output: {}'.format(output))
+    except subprocess.CalledProcessError as error:
+        if exit_code_indicates_error(error.returncode):
```
witten, line 242: Looks like this needs to pass in the `command`, too?

tests/unit/test_execute.py
```diff
@@ -239,6 +239,34 @@ def test_execute_command_and_capture_output_with_capture_stderr_returns_stderr()
...
+        full_command, stderr=None, shell=False, env=None, cwd=None
+    ).and_raise(subprocess.CalledProcessError(1, full_command, err_output)).once()
+    flexmock(module).should_receive('exit_code_indicates_error').and_return(False).once()
+
+    output = module.execute_command_and_capture_output(full_command)
+
+    assert output == expected_output
+
+
+def test_execute_command_and_capture_output_returns_output_when_error_code_not_one():
```
witten, line 257: Shouldn't this be more like `test_execute_command_and_capture_output_raises_when_command_errors()` or something?

PR #57, opened 2023-03-22
feat: tag repos

diivi:
Just a POC with only  `create_archive` accepting a repo label for now. 
Once reviewed, adding label support to other borgmatic `--repository` commands should be easier.

This is what it looks like:
image
with this config:
image

Tasks:
- [x] fix Merge conflicts (probably the format() and fstring thing).
- [x] two end to end tests failing for some reason.
- [x] Add normalization test
- [x] Add to docs

e2e:
image
unit:
image


borgmatic/config/normalize.py
```diff
@@ -56,9 +56,13 @@ def normalize(config_filename, config):
...
     # Upgrade remote repositories to ssh:// syntax, required in Borg 2.
     repositories = location.get('repositories')
+    if isinstance(repositories[0], str):
+        config['location']['repositories'] = [{'path': repository} for repository in repositories]
+        repositories = config['location']['repositories']
     if repositories:
         config['location']['repositories'] = []
-        for repository in repositories:
+        for repository_dict in repositories:
+            repository = repository_dict['path']
```
witten, line 65: For clarity, I'd suggest calling this variable `repository_path` or similar.

borgmatic/config/schema.yaml
```diff
@@ -29,19 +29,30 @@ properties:
...
-                    - /var/local/backups/local.borg
+                    type: object 
+                    required:
+                        - path
+                    properties:
+                        path:
+                            type: string
+                            description: |
+                                Path to local or remote repository (required). 
+                                are expanded. Multiple repositories are backed up to
```
witten, line 40: Missing start of sentence here!

borgmatic/config/schema.yaml
```diff
@@ -29,19 +29,30 @@ properties:
...
+                                in sequence. Borg placeholders can be used. See the
+                                output of "borg help placeholders" for details. See
+                                ssh_command for SSH options like identity file or
+                                port. If systemd service is used, then add local
+                                repository paths in the systemd service file to the
+                                ReadWritePaths list.                                
+                            example:
+                                - ssh://user@backupserver/./sourcehostname.borg
+                                - ssh://user@backupserver/./{fqdn}
+                                - /var/local/backups/local.borg
```
witten, line 50: Example should just be a string now.

borgmatic/config/schema.yaml
```diff
@@ -29,19 +29,30 @@ properties:
...
+                        label:
+                            type: string
+                            description: |
+                                Optional label for the repository.
```
witten, line 54: It might be good to suggest here what the label is used for.

borgmatic/config/validate.py
```diff
@@ -138,9 +138,15 @@ def normalize_repository_path(repository):
...
-    return normalize_repository_path(first) == normalize_repository_path(second)
+    Given two repository dicts with keys 'path' (relative and/or absolute),
+    and 'label', return whether they match.
+    '''
+    if isinstance(first, str) and isinstance(second, str):
+        return normalize_repository_path(first) == normalize_repository_path(second)
+    elif isinstance(first, dict) and isinstance(second, str):
+        return (second == first.get('label')) or (
+            normalize_repository_path(second) == normalize_repository_path(first.get('path'))
+        )
```
witten, line 149: If there are implicit expectations here about whether the second repository can be a dict or not, I'd recommend changing the function parameter names and docstring accordingly. (E.g., change `first` and `second` to `config_repository` and `repository_name` or something.) Because if I'm reading this right, if you pass in two dicts right now, this function won't work! An alternative would be to handle any `first` and `second` parameters regardless of whether they are dicts or strings.

witten (commented):
Awesome! This looks like a great approach.

diivi:
@witten can you help me understand why the full-dev-tests are failing? I attached the terminal output in the PR description.

diivi:
Nevermind, I dug deeper and was able to find the issue; even the examples matter in the schema file!

diivi:
Some dev workflow related questions:
1. can I format the `schema.yaml` file with a command?
2. can I run e2e tests only, when I run `scripts/run-full-dev-tests` or is there a script that will run them without running unit tests? Running unit tests seems simpler and faster with tox.

witten:
> Some dev workflow related questions:

>    can I format the schema.yaml file with a command?

Discussed on IRC. Spoiler for anyone else interested: Use your editor's/IDE's word wrapping functionality.

>    can I run e2e tests only, when I run scripts/run-full-dev-tests or is there a script that will run them without running unit tests? Running unit tests seems simpler and faster with tox.

I don't disagree, but the reason both are run is because `run-full-dev-tests` shares a scripting code path with CI tests that run on projects.torsion.org. And in the CI case, we definitely want both unit tests and end-to-end tests running. I'll see if there's a way to make this only run end-to-end tests in dev though.

witten:
>  I'll see if there's a way to make this only run end-to-end tests in dev though.

Done! `scripts/run-full-dev-tests` has been renamed to `scripts/run-end-to-end-dev-tests` and does exactly that. Full tests still run in CI. Thanks for the suggestion... I'm all about improving dev workflow.

borgmatic/config/schema.yaml
```diff
@@ -29,19 +29,40 @@ properties:
...
+                                If systemd service is used, then
+                                add local repository paths in the
+                                systemd service file to the
+                                ReadWritePaths list.
+                            example: ssh://user@backupserver/./{fqdn}
+                        label:
+                            type: string
+                            description: |
+                                Optional label for the repository. This 
+                                can be used with the --repository option
```
witten, line 56: "option" -> "flag" here. In user-facing stuff, I try to be pretty rigorous about using "option" to refer to configuration file values and "flag" to refer to command-line arguments. Cuts down on confusion, IMO.

borgmatic/config/validate.py
```diff
@@ -137,9 +137,17 @@ def normalize_repository_path(repository):
 
 def repositories_match(first, second):
     '''
-    Given two repository paths (relative and/or absolute), return whether they match.
-    '''
-    return normalize_repository_path(first) == normalize_repository_path(second)
+    Given two repository dicts with keys 'path' (relative and/or absolute),
+    and 'label', return whether they match.
```
witten, line 141: Maybe this should say: "Given two repository dicts or path strings ..."? Or something similar to suggest that it accepts either one.

borgmatic/config/validate.py
```diff
@@ -137,9 +137,17 @@ def normalize_repository_path(repository):
...
+    '''
+    if isinstance(first, str):
+        first = {'path': first, 'label': first}
+    if isinstance(second, str):
+        second = {'path': second, 'label': second}
+    return (first.get('label') == second.get('label')) or (
+        normalize_repository_path(first.get('path'))
+        == normalize_repository_path(second.get('path'))
+    )
```
witten, line 150: This is about as straightforward as this can be!

tests/integration/config/test_validate.py
```diff
@@ -63,7 +63,10 @@ def test_parse_configuration_transforms_file_into_mapping():
     config, logs = module.parse_configuration('/tmp/config.yaml', '/tmp/schema.yaml')
 
     assert config == {
-        'location': {'source_directories': ['/home', '/etc'], 'repositories': ['hostname.borg']},
+        'location': {
+            'source_directories': ['/home', '/etc'],
+            'repositories': [{'path': 'hostname.borg', 'label': ''}],
```
witten, line 68: Wouldn't it make more sense to omit `label` if it's not set? After all, it's optional. Similar throughout this test file.

tests/unit/config/test_normalize.py
```diff
@@ -69,27 +69,27 @@
         ),
         (
             {'location': {'repositories': ['foo@bar:/repo']}},
-            {'location': {'repositories': ['ssh://foo@bar/repo']}},
+            {'location': {'repositories': [{'path': 'ssh://foo@bar/repo', 'label': ''}]}},
```
witten, line 72: Similar thing here. (It probably would require changing what this function outputs though.)

witten (changes requested):
The overall approach looks good to me! This is an exciting feature to have, and I know I for one would start using it immediately. I left some comments with quibbles about minor details, hopefully all easy to fix. (Or easy to tell me why I'm wrong about them.)

I didn't do an exhaustive search of the source to see if there were any conversions from repository str to dict missed. Unit tests would catch some of that, but things could still slip through the cracks especially if an omission happens _between_ units. My recommendation would be to try running through several borgmatic actions if you haven't already, just to see if any of them blow up.

witten:
Also, feel free to update repository examples in the README and other documentation for this if you want. In any case, I certainly don't mind writing up documentation for this feature in particular.

borgmatic/config/normalize.py
```diff
@@ -57,9 +57,15 @@ def normalize(config_filename, config):
     # Upgrade remote repositories to ssh:// syntax, required in Borg 2.
     repositories = location.get('repositories')
     if repositories:
+        if isinstance(repositories[0], str):
+            config['location']['repositories'] = [
+                {'path': repository} for repository in repositories
+            ]
```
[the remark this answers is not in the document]
diivi: This doesn't support a mix and match between repository dicts and strs, so might need to document that (note to self).

borgmatic/commands/borgmatic.py
```diff
@@ -259,13 +261,15 @@ def run_actions(
     invalid.
     '''
     add_custom_log_levels()
-    repository = os.path.expanduser(repository_path)
+    if isinstance(repository, str):
+        repository = {'path': repository}
```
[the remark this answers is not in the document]
diivi: Ah, looking back now, I probably changed these to pass the tests; I should've changed the tests though.
(The tests after the normalization phase that are sending strings to these functions).

tests/unit/commands/test_borgmatic.py
```diff
@@ -249,41 +249,41 @@ def test_run_configuration_repos_ordered():
...
         'bar: Error running actions for repository', OSError
     ).and_return(expected_results[1:]).ordered()
-    config = {'location': {'repositories': ['foo', 'bar']}}
+    config = {'location': {'repositories': [{'path': 'foo'}, {'path': 'bar'}]}}
     arguments = {'global': flexmock(monitoring_verbosity=1, dry_run=False), 'create': flexmock()}
     results = list(module.run_configuration('test.yaml', config, arguments))
     assert results == expected_results
 
 
-def test_run_configuration_retries_round_robbin():
```
diivi, line 258: I can't get this one test to pass, something to do with the order these functions are called in:
```bash
flexmock.CallOrderError:
log_error_records("bar: Error running actions for repository", , levelno=30, log_command_error_output=True) 
called before
log_error_records(levelno=30, log_command_error_output=True, message="foo: Error running actions for repository", error=<class 'OSError'>)
```

Any suggestion on what would be the right way to fix this?

borgmatic/actions/rlist.py
```diff
@@ -19,10 +19,10 @@ def run_rlist(
         repository, rlist_arguments.repository
     ):
         if not rlist_arguments.json:  # pragma: nocover
-            logger.answer(f'{repository}: Listing repository')
+            logger.answer(f'{repository["path"]}: Listing repository')
 
         json_output = borgmatic.borg.rlist.list_repository(
-            repository,
+            repository['path'],
```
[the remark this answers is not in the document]
diivi: will do this in one commit, so it's easier to check.

tests/unit/commands/test_borgmatic.py
witten: > Any suggestion on what would be the right way to fix this?

I didn't write this test, but I think what's going on is the config setup is incorrect:

Simply change this:

```python
config = {'location': {'repositories': [{'path':'foo','path':'bar'}]}, 'storage': {'retries': 1}}
```

... to this:

```python
config = {'location': {'repositories': [{'path':'foo'}, {'path':'bar'}]}, 'storage': {'retries': 1}}
```

> (Also this should probably be round_robin)

Yes! Good catch.

diivi:
Where in the docs should this be covered?🤔

witten:
> Where in the docs should this be covered?

I don't think that this necessarily needs its own dedicated section; the feature could just be mentioned whenever `--repository` is mentioned in the docs. E.g., let the user know they can pass in _either_ a configured repository path or a repository label at that point. Right now, such pages include:

 * https://torsion.org/borgmatic/docs/how-to/extract-a-backup/
 * https://torsion.org/borgmatic/docs/how-to/backup-your-databases/
 * https://torsion.org/borgmatic/docs/how-to/run-arbitrary-borg-commands/

> While browsing the docs, I got another idea - maybe a search bar for the docs like the one borg has would be useful? I can try doing some research and let you know if it's possible.

That would be awesome, but it may be a bit of work. I think the search term for that search functionality is "11ty search." I'd also be open to reorganizing the docs if their current structure is making it hard to find stuff.

witten (approved):
This looks great! Left one minor comment, but it's scope creep and not enough to block approval on. Nice work!

PR #58, opened 2023-03-23
docs: copy to clipboard support

diivi:
Adds a quick copy to clipboard button to every codeblock in the docs.
Uses https://github.com/mamezou-tech/eleventy-plugin-code-clipboard

image

Let me know what you think!

### DIFF
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -18,10 +18,12 @@
 {% if feedTitle and feedUrl %}
 		<link rel="alternate" href="{{ feedUrl }}" title="{{ feedTitle }}" type="application/atom+xml">
 {% endif %}
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@mdi/font@6.5.95/css/materialdesignicons.min.css" crossorigin="anonymous" referrerpolicy="no-referrer" />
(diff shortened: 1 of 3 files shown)


docs/_includes/layouts/base.njk
```diff
@@ -18,10 +18,12 @@
 {% if feedTitle and feedUrl %}
 		<link rel="alternate" href="{{ feedUrl }}" title="{{ feedTitle }}" type="application/atom+xml">
 {% endif %}
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@mdi/font@6.5.95/css/materialdesignicons.min.css" crossorigin="anonymous" referrerpolicy="no-referrer" />
```
witten, line 21: I'll admit I'm not wild about pulling down 300k just for a single icon, but I guess it'll be cached after the first page load. (Or even before then, if it's on a common CDN.)

witten:
This is amazing! Thanks for taking the initiative to just do it.

docs/_includes/layouts/base.njk
diivi: Good point, it didn't cross my mind when I was trying to get it to work 😅. Another dev raised the same point here - https://github.com/mamezou-tech/eleventy-plugin-code-clipboard/issues/8.
Apparently, both these css links can be removed, and it'll still work (I wasn't using the tooltips anyway). I got it to work, will submit a patch right away!

witten: Awesome!

PR #59, opened 2023-03-23
fix: remove extra links from docs css

diivi:
Improve page load time (or just remove extra links).

Works the same.


witten:
Thank you!!

PR #60, opened 2023-03-23
feat: allow defining custom variables in config file

diivi:
- [x]  Tests pass
- [x] can replace primitive constants
- [x] Support more [complex](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/612#issuecomment-5715) data types.

Still have to add support for the third point from `generate.py::render_configuration()` and test it out.

Questions:
1. Should this support more complex cases like these:

```yaml
constants:
  x: /home/divi/folder
  y: {x}/file.txt
```

### DIFF
--- a/borgmatic/config/load.py
+++ b/borgmatic/config/load.py
@@ -1,4 +1,5 @@
 import functools
+import json
 import logging
 import os
 
@@ -104,7 +105,8 @@ def load_configuration(filename):
         config = yaml.load(file_contents)
         if config and 'constants' in config:
             for key, value in config['constants'].items():
-                file_contents = file_contents.replace(f'{{{key}}}', str(value))
+                value = json.dumps(value)
+                file_contents = file_contents.replace(f'{{{key}}}', value)
             config = yaml.load(file_contents)
             del config['constants']
         return config
--- a/borgmatic/config/load.py
+++ b/borgmatic/config/load.py
@@ -100,8 +100,8 @@ def load_configuration(filename):
     yaml = ruamel.yaml.YAML(typ='safe')
     yaml.Constructor = Include_constructor_with_include_directory
 
-    with open(filename) as f:
-        file_contents = f.read()
+    with open(filename) as file:
+        file_contents = file.read()
         config = yaml.load(file_contents)
         if config and 'constants' in config:
             for key, value in config['constants'].items():
--- a/borgmatic/config/load.py
+++ b/borgmatic/config/load.py
@@ -81,7 +81,8 @@ class Include_constructor(ruamel.yaml.SafeConstructor):
 def load_configuration(filename):
     '''
     Load the given configuration file and return its contents as a data structure of nested dicts
-    and lists.
+    and lists. Also, replace any "{constant}" strings with the value of the "constant" key in the
+    "constants" section of the configuration file.
 
     Raise ruamel.yaml.error.YAMLError if something goes wrong parsing the YAML, or RecursionError
     if there are too many recursive includes.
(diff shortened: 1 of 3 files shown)


witten:
> Questions:
>
>    Should this support more complex cases like these:
>
> ```
>  constants:
>   x: /home/divi/folder
>   y: {x}/file.txt
> ```

Ugh. On the one hand, I can see someone actually wanting to construct multiple variables like this to save on typing. On the other hand, we're not building a programming language here. And it'd be a pain to implement, too, as you'd presumably have to do n passes of re-loading the file where n is the number of constants! So I'm totally okay with *not* supporting this, documenting it as a very intentional limitation, and then waiting for user feedback to see if anyone's clamoring for this feature.

borgmatic/config/load.py
```diff
@@ -98,7 +99,15 @@ def __init__(self, preserve_quotes=None, loader=None):
     yaml = ruamel.yaml.YAML(typ='safe')
     yaml.Constructor = Include_constructor_with_include_directory
 
-    return yaml.load(open(filename))
+    with open(filename) as f:
```
witten, line 102: Nit: No single-letter variables. (I realize this is kind of obnoxious. :smile:)

witten (commented):
This looks great!!!

borgmatic/config/load.py
diivi: Ah, I keep forgetting 😅. Is there maybe a flake8 rule or something for this? The consistency makes sense to me. And we might add one for using `f"` strings instead of `.format`  if that's consistent too.

diivi: found this - https://pypi.org/project/flake8-variables-names/ and this - https://github.com/MichaelKim0407/flake8-use-fstring. They don't seem to be that popular though.

witten: Yes, I was just looking at them too after your first comment! I'll integrate one or both of these or update https://torsion.org/borgmatic/docs/how-to/develop-on-borgmatic/ page to mention these conventions.

diivi: That was a lot of f'strings indeed 😆!

diivi:
Just realised, I can't  really do what you asked me to do in the discussion (refer to `render_configuration()`). Because while using yaml.dump, I will get an error (Probably because {x} is not valid syntax in yaml):
```bash
did not find expected <document start>
  in "<unicode string>", line 22, column 5
```

So for complex data types I would probably have to do something complex in code too?

diivi:
I am having trouble calculating the indentation for replacement of complex constants.

`dumper.indent(mapping=2, sequence=4, offset=2)`

This was for the complete config I think, but for one particular constant, how will I know these values ( how deep it is supposed to be nested)?

witten:
I may have steered you in the wrong direction—for exactly the indentation reasons you're discovering! So how about a different approach? Here are two things I know:

 * The configuration you're "generating" is not for user consumption, so it doesn't need to be pretty.
 * YAML is a superset of JSON, meaning that you can use JSON directly as values in YAML.

So how about instead of dumping YAML (and having to worry indentation), you just dump a single line of JSON? That should handle dicts, lists, etc.

witten:
Beautiful! Minor code style thing for future reference: I usually put a blank line before `if` statements just for general readability.

witten:
I actually found a problem with this feature due to the JSON approach, and I'm not sure what to do about it. Doing a simple string substitution like this ...

```yaml
constants:
    foo: tmp

location:
    source_directories:
        /root/{foo}
```

... results in a replaced string that looks like this ...

```yaml
location:
    source_directories:
        /root/"tmp"
```

Which I don't think is what we want, especially since single string substitution is probably the most common use case!

witten:
Could a fix be as simple as calling `.strip('"')` after the `json.loads(...)`? I don't think that would hurt any other use cases (integers, lists, dicts, etc.).

diivi:
Maybe I should do a json dump replacement only when a constant is an instance of dict/list? Otherwise, the normal str(value) replacement I was doing earlier.

diivi:
> Could a fix be as simple as calling `.strip('"')` after the `json.loads(...)`? I don't think that would hurt any other use cases (integers, lists, dicts, etc.).

Let me try this and get back to you.

witten:
Either approach would work! Your idea might even be less hacky.

PR #61, opened 2023-03-24
fix: docs cli reference create spelling

diivi:
Was checking out [#462](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/462), trying to find out where the default behaviour for `borgmatic` is documented.
Found this small spelling mistake.
This should change the [docs](https://torsion.org/borgmatic/docs/reference/command-line/) too right?


witten:
Good eye! Thanks for fixing this. Yes, this will update the command-line reference docs, too!

PR #63, opened 2023-03-28
fix: make check repositories work with dict and str repositories

diivi:
[#662](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/662)

### DIFF
--- a/tests/unit/config/test_validate.py
+++ b/tests/unit/config/test_validate.py
@@ -51,13 +51,29 @@ def test_apply_locical_validation_raises_if_unknown_repository_in_check_reposito
         )
 
 
-def test_apply_locical_validation_does_not_raise_if_known_repository_in_check_repositories():
+def test_apply_locical_validation_does_not_raise_if_known_repository_path_in_check_repositories():
     module.apply_logical_validation(
         'config.yaml',
         {
-            'location': {'repositories': ['repo.borg', 'other.borg']},
+            'location': {'repositories': [{'path': 'repo.borg'}, {'path': 'other.borg'}]},
             'retention': {'keep_secondly': 1000},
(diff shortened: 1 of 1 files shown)


witten:
I think this is it! (At some point it might make sense to move this into the repositories list, although I'm trying to avoid nesting an entire config file under `repositories:`)

Are there any tests that can/should be updated along with this fix?

witten (commented):
Looks good.. Nice use of `repositories_match`! See my question about potential tests though.

diivi:
nothing to change for the `tox` tests; I am getting this error when I run the end to end tests though:


image

witten:
> nothing to change for the tox tests

What I was hoping for is: Is there a way to make a failing unit (or integration) test in master, apply your fix, and observe that the failing test now passes? Just something to give us more confidence in the fix.. I totally agree though that just _looking_ at your change, it does appear correct.

> I am getting this error when I run the end to end tests though:

Huh, super weird. End-to-end tests run just fine here. It's looking like `apk` isn't found in your container image...? I have no idea what might cause that. You didn't accidentally switch the image from Alpine to something else, did you? :smile: But more seriously, any other local changes that might impact this?

diivi:
> any other local changes that might impact this?

I don't really think so, apart from updating the branch and running `scripts/run-full-tests` for the first time instead of the earlier `dev-tests`, I haven't changed anything.

witten:
Oh, try running `run-end-to-end-dev-tests` instead! I should probably make it so users can run `run-full-tests` by itself. (See the comments at the top.)

witten:
Looks great, thank you!

witten:
FYI I've added a check to `run-full-tests` that errors out if it's run manually (not in a test environment).

PR #66, opened 2023-04-03
add favicon to documentation

diivi:
I thought this was missing 😅.
A green/grey icon would be better for dark mode users, though.
image

Before:
image

After:
image

(Irrelevant, but why is the title on the home page not capitalised?)

### DIFF
--- a/docs/_includes/layouts/base.njk
+++ b/docs/_includes/layouts/base.njk
@@ -3,6 +3,7 @@
 	<head>
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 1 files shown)


witten:
Awesome! I really appreciate getting these fit and finish fixes. Side note: I think there might be a way to do [dark mode favicon](https://css-tricks.com/dark-mode-favicons/), but maybe only if the icon is [SVG](https://icons8.com/icons/set/borg).

> (Irrelevant, but why is the title on the home page not capitalised?)

To avoid a disparity between borgmatic the project and borgmatic the command-line program, I named the project "borgmatic", all lowercase. This is something I've seen in other command-line projects as well. Although judging by how often the project is called "Borgmatic", maybe it wasn't the best idea?

PR #67, opened 2023-04-05
feat: restore specific schemas

diivi:
Fixes [#375](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/375)

- [x] Restore a specific psql schema.
- [x] Restore multiple psql schemas at the same time. For this, maybe I have to use multiple `-n`s instead of `--schema`. Ref: the schema section here - https://www.postgresql.org/docs/current/app-pgrestore.html 
- [x] Restore multiple MongoDB collections at the same time.

### DIFF
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -68,12 +68,15 @@ def restore_single_database(
     archive_name,
     hook_name,
     database,
+    schemas
 ):  # pragma: no cover
     '''
     Given (among other things) an archive name, a database hook name, and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(f'{repository}: Restoring database {database["name"]}')
+    if schemas:
+        database['schemas'] = schemas
 
     dump_pattern = borgmatic.hooks.dispatch.call_hooks(
         'make_database_dump_pattern',
@@ -314,6 +317,7 @@ def run_restore(
                 archive_name,
                 found_hook_name or hook_name,
                 found_database,
+                schemas = restore_arguments.schemas,
             )
 
     # For any database that weren't found via exact matches in the hooks configuration, try to
@@ -343,6 +347,7 @@ def run_restore(
                 archive_name,
                 found_hook_name or hook_name,
                 database,
+                schemas = restore_arguments.schemas,
             )
 
     borgmatic.hooks.dispatch.call_hooks_even_if_unconfigured(
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -68,15 +68,12 @@ def restore_single_database(
     archive_name,
     hook_name,
     database,
-    schemas
 ):  # pragma: no cover
     '''
     Given (among other things) an archive name, a database hook name, and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(f'{repository}: Restoring database {database["name"]}')
-    if schemas:
-        database['schemas'] = schemas
 
     dump_pattern = borgmatic.hooks.dispatch.call_hooks(
         'make_database_dump_pattern',
@@ -316,8 +313,7 @@ def run_restore(
                 remote_path,
                 archive_name,
                 found_hook_name or hook_name,
-                found_database,
-                schemas = restore_arguments.schemas,
+                dict(found_database, **{'schemas': restore_arguments.schemas}),
             )
 
     # For any database that weren't found via exact matches in the hooks configuration, try to
@@ -346,8 +342,7 @@ def run_restore(
                 remote_path,
                 archive_name,
                 found_hook_name or hook_name,
-                database,
-                schemas = restore_arguments.schemas,
+                dict(database, **{'schemas': restore_arguments.schemas}),
             )
 
     borgmatic.hooks.dispatch.call_hooks_even_if_unconfigured(
--- a/borgmatic/commands/arguments.py
+++ b/borgmatic/commands/arguments.py
@@ -634,7 +634,7 @@ def make_parsers():
         metavar='NAME',
         nargs='+',
         dest='schemas',
-        help="Names of schemas to restore from the database, defaults to all schemas."
+        help="Names of schemas to restore from the database, defaults to all schemas. Schemas are only supported for PostgreSQL and MongoDB databases",
     )
     restore_group.add_argument(
         '-h', '--help', action='help', help='Show this help message and exit'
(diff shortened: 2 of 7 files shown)


borgmatic/hooks/postgresql.py
```diff
@@ -213,6 +213,7 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         + ('--command', 'ANALYZE')
     )
     pg_restore_command = database.get('pg_restore_command') or 'pg_restore'
+    backup_schemas = ', '.join(database['schemas']) if 'schemas' in database else None
```
diivi, line 216: This is not how it works 😅.

witten:
> maybe I have to use multiple -ns instead of --schema

If I'm understanding the docs correctly, that's not necessarily the case. I think they're just saying you'd have to do `-n schema1 -n schema2` *or* `--schema=schema1 --schema=schema2`. And FYI you can do that sort of variable arguments support with borgmatic too if you want. E.g., look at the existing `--config` option for an example.

borgmatic/actions/restore.py
```diff
@@ -68,12 +68,15 @@ def restore_single_database(
...
     database,
+    schemas
 ):  # pragma: no cover
     '''
     Given (among other things) an archive name, a database hook name, and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(f'{repository}: Restoring database {database["name"]}')
+    if schemas:
+        database['schemas'] = schemas
```
witten, line 79: There's a risk that this mutates the database dict and "leaks" that mutation to callers of `restore_single_database`. Consider making a copy of `database` and mutating that instead. Alternatively, since it looks like the whole `database` dict is only used in a single place below, you could even do it inline:

```python
dict(database, **{'schema': schema})
```

(That syntax sidesteps duplicate keyword collisions on `schema`.)

borgmatic/actions/restore.py
```diff
@@ -314,6 +317,7 @@ def run_restore(
                 archive_name,
                 found_hook_name or hook_name,
                 found_database,
+                schemas = restore_arguments.schemas,
```
witten, line 320: Nit: No spaces around the `=` operator here. (Flake8 might catch this.)

borgmatic/commands/arguments.py
```diff
@@ -629,6 +629,13 @@ def make_parsers():
         dest='databases',
         help="Names of databases to restore from archive, defaults to all databases. Note that any databases to restore must be defined in borgmatic's configuration",
     )
+    restore_group.add_argument(
+        '--schema',
+        metavar='NAME',
+        nargs='+',
+        dest='schemas',
+        help="Names of schemas to restore from the database, defaults to all schemas."
```
witten, line 637: Maybe worth noting "if supported by your database system" or whatever? Or even specifying the database hooks that support this?

borgmatic/commands/arguments.py
```diff
@@ -629,6 +629,13 @@ def make_parsers():
         dest='databases',
         help="Names of databases to restore from archive, defaults to all databases. Note that any databases to restore must be defined in borgmatic's configuration",
     )
+    restore_group.add_argument(
+        '--schema',
+        metavar='NAME',
+        nargs='+',
```
witten, line 635: Looks like you already knew about `nargs`! Nice.

borgmatic/hooks/postgresql.py
witten: :smile:

witten (commented):
Your approach looks good to me! No major feedback.

borgmatic/hooks/postgresql.py
```diff
@@ -226,6 +226,11 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         + (tuple(database['restore_options'].split(' ')) if 'restore_options' in database else ())
         + (() if extract_process else (dump_filename,))
     )
+
+    if database['schemas']:
+        for schema in database['schemas']:
+            restore_command += ('--schema', schema)
```
witten, line 232: If you wanted to, this could be worked into the `restore_command` declaration above instead of extending the command here. Something like:

```python
  + tuple(itertools.chain.from_iterable(('--schema', schema) for schema in database['schemas']))
```

(Do not feel strongly.)

borgmatic/actions/restore.py
```diff
@@ -313,7 +313,7 @@ def run_restore(
                 remote_path,
                 archive_name,
                 found_hook_name or hook_name,
-                found_database,
+                dict(found_database, **{'schemas': restore_arguments.schemas}),
```
witten, line 316: Clever way to pass this in!

borgmatic/hooks/mongodb.py
```diff
@@ -161,4 +161,7 @@ def build_restore_command(extract_process, database, dump_filename):
         command.extend(('--authenticationDatabase', database['authentication_database']))
     if 'restore_options' in database:
         command.extend(database['restore_options'].split(' '))
+    if database['schemas']:
+        for schema in database['schemas']:
+            command.extend(('--nsInclude', schema))
```
witten, line 166: Would it be possible to add unit tests for the `schemas` case with these MongoDB and PostgreSQL functions? In the case of PostgreSQL, the simplest thing would be to add another test for `restore_database_dump()` with appropriate mocking. An alternative though would to be to break out a function that builds the restore command and then test _that_. However, that would require more changes to existing tests. So, a trade-off either way.

witten (changes requested):
The code looks good to me. The only thing missing to my eye is some basic tests for this new functionality. Thanks!

borgmatic/hooks/postgresql.py
diivi: Nice!

borgmatic/hooks/postgresql.py
```diff
@@ -225,7 +226,13 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
         + (('--username', database['username']) if 'username' in database else ())
         + (tuple(database['restore_options'].split(' ')) if 'restore_options' in database else ())
         + (() if extract_process else (dump_filename,))
+        + tuple(
+            itertools.chain.from_iterable(('--schema', schema) for schema in database['schemas'])
+            if database['schemas']
```
witten, line 231: Good call on including the `if` here!

witten (approved):
Awesome! LGTM.

witten:
Thanks for tackling this one!

PR #68, opened 2023-04-06
feat: add logfile name to hook context for interpolation

diivi:
... when --log-file is passed as an argument

Should Fix [#413](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/413)

### DIFF
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -269,6 +269,12 @@ def run_actions(
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
     }
 
+    try:
+        log_file = global_arguments.log_file
+        hook_context['log_file'] = log_file
+    except AttributeError:
+        pass
+
     command.execute_hook(
         hooks.get('before_actions'),
         hooks.get('umask'),
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -267,9 +267,8 @@ def run_actions(
         'repository': repository_path,
         # Deprecated: For backwards compatibility with borgmatic < 1.6.0.
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
+        'log_file': global_arguments.log_file if global_arguments.log_file else '',
     }
-
-    hook_context['log_file'] = global_arguments.log_file if global_arguments.log_file else ''
 
     command.execute_hook(
         hooks.get('before_actions'),
--- a/borgmatic/commands/borgmatic.py
+++ b/borgmatic/commands/borgmatic.py
@@ -269,11 +269,7 @@ def run_actions(
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
     }
 
-    try:
-        log_file = global_arguments.log_file
-        hook_context['log_file'] = log_file
-    except AttributeError:
-        pass
+    hook_context['log_file'] = global_arguments.log_file if global_arguments.log_file else ''
 
     command.execute_hook(
         hooks.get('before_actions'),
--- a/tests/unit/commands/test_borgmatic.py
+++ b/tests/unit/commands/test_borgmatic.py
@@ -422,6 +422,26 @@ def test_run_actions_runs_rcreate():
         )
     )
 
+def test_run_actions_adds_log_file_to_hook_context():
+    flexmock(module).should_receive('add_custom_log_levels')
+    flexmock(module.command).should_receive('execute_hook')
+    flexmock(borgmatic.actions.rcreate).should_receive('run_rcreate').once()
+
+    tuple(
+        module.run_actions(
+            arguments={'global': flexmock(dry_run=False, log_file='foo'), 'rcreate': flexmock()},
+            config_filename=flexmock(),
+            location={'repositories': []},
+            storage=flexmock(),
+            retention=flexmock(),
+            consistency=flexmock(),
+            hooks={},
+            local_path=flexmock(),
+            remote_path=flexmock(),
+            local_borg_version=flexmock(),
+            repository={'path': 'repo'},
+        )
+    )
 
 def test_run_actions_runs_transfer():
     flexmock(module).should_receive('add_custom_log_levels')
--- a/tests/unit/commands/test_borgmatic.py
+++ b/tests/unit/commands/test_borgmatic.py
@@ -422,25 +422,27 @@ def test_run_actions_runs_rcreate():
         )
     )
 
+
 def test_run_actions_adds_log_file_to_hook_context():
     flexmock(module).should_receive('add_custom_log_levels')
     flexmock(module.command).should_receive('execute_hook')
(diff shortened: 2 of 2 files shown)


borgmatic/commands/borgmatic.py
```diff
@@ -269,6 +269,12 @@ def run_actions(
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
     }
 
+    try:
+        log_file = global_arguments.log_file
+        hook_context['log_file'] = log_file
+    except AttributeError:
```
witten, line 275: I don't think this will actually ever raise because `log_file` is always an attribute on `global_arguments` even if the value is `None`. So an `if` instead of a `try`/`except` might work just fine. Although.. I'd be fine unconditionally setting this into `hook_context` when it's originally assigned above, maybe empty string when the `log_file` value is `None`...?

borgmatic/commands/borgmatic.py
```diff
@@ -269,6 +269,12 @@ def run_actions(
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
     }
 
+    try:
+        log_file = global_arguments.log_file
+        hook_context['log_file'] = log_file
```
witten, line 274: This two lines could be combined into one without the intermediate variable.

tests/unit/commands/test_borgmatic.py
```diff
@@ -422,6 +422,26 @@ def test_run_actions_runs_rcreate():
         )
     )
 
+def test_run_actions_adds_log_file_to_hook_context():
```
witten, line 425: I'm not seeing any asserts/expectations that log file is actually making its way into `hook_context` though..?

witten (commented):
Ooh thanks for grabbing this one.

borgmatic/commands/borgmatic.py
diivi: Ah, I got the `AttributeError` from tests, but again, I should've updated the tests only.

tests/unit/commands/test_borgmatic.py
```diff
@@ -422,6 +422,39 @@ def test_run_actions_runs_rcreate():
...
+    flexmock(module).should_receive('add_custom_log_levels')
+    flexmock(module.command).should_receive('execute_hook')
+    flexmock(borgmatic.actions.create).should_receive('run_create').with_args(
```
diivi, line 428: Can you help me understand why this does not work?
Flexmock throws this error:
```bash
E           flexmock.MethodSignatureError: run_create(<flexmock.MockClass object at 0x7fc8afc78340>, {'path': 'repo'}, {'repositories': []}, <flexmock.MockClass object at 0x7fc8afc4bb80>, {}, {'repository': 'repo', 'repositories': '', 'log_file': 'foo'}, <flexmock.MockClass object at 0x7fc8afc4bb50>, <flexmock.MockClass object at 0x7fc8afc78220>, <flexmock.MockClass object at 0x7fc8afc48cd0>, "", <flexmock.MockClass object at 0x7fc8afc4a050>, <flexmock.MockClass object at 0x7fc8afc49990>)
```
But isn't the method signature correct?

witten: The first discrepancy I see is the `hook_context`. You're asserting that the `hook_context` is `{'log_file': 'foo'}`, but the Flexmock error indicates that it's actually `{}`.

borgmatic/commands/borgmatic.py
```diff
@@ -269,6 +269,8 @@ def run_actions(
         'repositories': ','.join([repo['path'] for repo in location['repositories']]),
     }
 
+    hook_context['log_file'] = global_arguments.log_file if global_arguments.log_file else ''
```
witten, line 272: You could fold this directly into the `hook_context` declaration if you wanted. (Less mutation.) Example:

```python
hook_context = {
   ...
   'repositories': ...
   'log_file': global_arguments.log_file if global_arguments.log_file else ''
}
```

I do not feel strongly about this though.

tests/unit/commands/test_borgmatic.py
```diff
@@ -422,6 +422,39 @@ def test_run_actions_runs_rcreate():
...
+        storage=flexmock(),
+        hooks={},
+        hook_context={'log_file': 'foo'},
+        local_borg_version=flexmock(),
+        create_arguments=flexmock(),
+        global_arguments=flexmock(dry_run=False, log_file='foo'),
+        dry_run_label='',
+        local_path=flexmock(),
+        remote_path=flexmock(),
+    ).once()
```
witten, line 441: FYI there's a cool trick you can do with Flexmock if there's an argument you don't really care about setting an expectation for. Just set its value inside a `.with_args()` call to `object`. That effectively tells flexmock "any object will do here; I don't really care what the value is".

I'm not saying you have to do that here. Just letting you know because it often saves on boilerplate.

witten (commented):
The overall approach looks good to me!

tests/unit/commands/test_borgmatic.py
diivi: Thanks, I changed it to this:
```python
 flexmock(borgmatic.actions.create).should_receive('run_create').with_args(
        config_filename=flexmock(),
        repository={'path': 'repo'},
        location={'repositories': []},
        storage=flexmock(),
        hooks={},
        hook_context={'repository': 'repo', 'repositories': '', 'log_file': 'foo'},
        local_borg_version=flexmock(),
        create_arguments=flexmock(),
        global_arguments=flexmock(dry_run=False, log_file='foo'),
        dry_run_label='',
        local_path=flexmock(),
        remote_path=flexmock(),
    ).once()
```
But it still raises the  same error:

```bash
flexmock.MethodSignatureError: run_create(<flexmock.MockClass object at 0x7f1f465ce5c0>, {'path': 'repo'}, {'repositories': []}, <flexmock.MockClass object at 0x7f1f465cfd90>, {}, {'repository': 'repo', 'repositories': '', 'log_file': 'foo'}, <flexmock.MockClass object at 0x7f1f46435a80>, <flexmock.MockClass object at 0x7f1f465ce7a0>, <flexmock.MockClass object at 0x7f1f46458370>, "", <flexmock.MockClass object at 0x7f1f465cea40>, <flexmock.MockClass object at 0x7f1f465cee60>)
```

witten: I think what's going on is that the instances of `flexmock()` are not matching up. For instance, you'd need the `flexmock()` instance used in `run_actions()` to be the exact same instance used in your expectation here. If you don't care about that level of fidelity, then just use `object` instead to indicate that you expect _something_ passed in but you don't care what it is.

witten (approved):
Looks good to me. Thanks for working through this!

witten:
I also added this to the docs: https://projects.torsion.org/borgmatic-collective/borgmatic/commit/4cd7556a34f1cff6c8c87047f18e6b6872567ab9

PR #69, opened 2023-04-14
Fix PostgreSQL hook not using "psql_command" for list when dumping "all"

jirutka:


jirutka:
There are more bugs in `*_command` parameters: all but `dump_command` require a command without arguments, so these `docker exec` examples don’t work either.

```
Error running actions for repository
[Errno 2] No such file or directory: 'doas -u postgres psql'
```

jirutka:
Moved to https://projects.torsion.org/borgmatic-collective/borgmatic/pulls/677

PR #70, opened 2023-04-28
add fish completions support

isaec:
closes https://projects.torsion.org/borgmatic-collective/borgmatic/issues/686

I've tested this on my system and it seems to work well. There is room to improve on the completions, but parsing the args is a little tedious, and this works well enough for my use.

I don't write python, so I'm sorry if this isn't pythonic.

### DIFF
--- a/borgmatic/commands/completion.py
+++ b/borgmatic/commands/completion.py
@@ -198,11 +198,13 @@ def fish_completion():
             f'''complete -c borgmatic -f -n "$exact_option_condition" -a '{' '.join(action.option_strings)}' -d {shlex.quote(action.help)}{exact_options_completion(action)}'''
             for action in top_level_parser._actions
             if len(action.option_strings) > 0
+            if 'Deprecated' not in action.help
         )
         + ('\n# subparser flags',)
         + tuple(
             f'''complete -c borgmatic -f -n "$exact_option_condition" -a '{' '.join(action.option_strings)}' -d {shlex.quote(action.help)} -n "__fish_seen_subcommand_from {action_name}"{exact_options_completion(action)}'''
             for action_name, subparser in subparsers.choices.items()
             for action in subparser._actions
+            if 'Deprecated' not in action.help
         )
     )
--- a/borgmatic/commands/completion.py
+++ b/borgmatic/commands/completion.py
@@ -1,12 +1,13 @@
 from borgmatic.commands import arguments
 
-UPGRADE_MESSAGE = '''
-Your bash completions script is from a different version of borgmatic than is
+def upgrade_message(language: str, upgrade_command: str, completion_file: str):
+    return f'''
+Your {language} completions script is from a different version of borgmatic than is
 currently installed. Please upgrade your script so your completions match the
 command-line flags in your installed borgmatic! Try this to upgrade:
 
-    sudo sh -c "borgmatic --bash-completion > $BASH_SOURCE"
-    source $BASH_SOURCE
+    {upgrade_command}
+    source {completion_file}
 '''
 
 
@@ -34,7 +35,7 @@ def bash_completion():
             '    local this_script="$(cat "$BASH_SOURCE" 2> /dev/null)"',
             '    local installed_script="$(borgmatic --bash-completion 2> /dev/null)"',
             '    if [ "$this_script" != "$installed_script" ] && [ "$installed_script" != "" ];'
-            f'        then cat << EOF\n{UPGRADE_MESSAGE}\nEOF',
+            '        then cat << EOF\n{}\nEOF'.format(upgrade_message('bash', 'sudo sh -c "borgmatic --bash-completion > $BASH_SOURCE"', '$BASH_SOURCE')),
             '    fi',
             '}',
             'complete_borgmatic() {',
@@ -69,11 +70,11 @@ def fish_completion():
     return '\n'.join(
         (
             'function __borgmatic_check_version',
-            '    set this_script (cat "$BASH_SOURCE" 2> /dev/null)',
-            '    set installed_script (borgmatic --bash-completion 2> /dev/null)',
+            '    set this_script (status current-filename)',
+            '    set installed_script (borgmatic --fish-completion 2> /dev/null)',
             '    if [ "$this_script" != "$installed_script" ] && [ "$installed_script" != "" ]',
-            f'        echo "{UPGRADE_MESSAGE}"',
+            '        echo "{}"'.format(upgrade_message('fish', 'borgmatic --fish-completion | sudo tee (status current-filename)', '(status current-filename)')),
             '    end',
             'end',
-            'function __borgmatic_complete',
+            # 'function __borgmatic_complete',
         ))
--- a/borgmatic/commands/completion.py
+++ b/borgmatic/commands/completion.py
@@ -80,25 +80,24 @@ def fish_completion():
             'end',
             '__borgmatic_check_version &',
         ) + (
-            '# subparser completions',
+            '\n# subparser completions',
         ) + tuple(
             '''complete -c borgmatic -a '%s' -d %s -f -n "not __fish_seen_subcommand_from %s"'''
             % (actionStr, shlex.quote(subparser.description), all_subparsers)
             for actionStr, subparser in subparsers.choices.items()
         ) + (
-            '# global flags',
+            '\n# global flags',
         ) + tuple(
             '''complete -c borgmatic -a '%s' -d %s -f'''
             % (option, shlex.quote(action.help))
             for action in top_level_parser._actions
             for option in action.option_strings
         ) + (
-            '# subparser flags',
+            '\n# subparser flags',
         ) + tuple(
-            '''complete -c borgmatic -a '%s' -d %s -f -n "__fish_seen_subcommand_from %s"'''
-            % (option, shlex.quote(action.help), actionStr)
+            '''complete -c borgmatic -a '%s' -d %s -n "__fish_seen_subcommand_from %s" -f'''
+            % (' '.join(action.option_strings), shlex.quote(action.help), actionStr)
             for actionStr, subparser in subparsers.choices.items()
             for action in subparser._actions
-            for option in action.option_strings
         )
     )
--- a/borgmatic/commands/completion.py
+++ b/borgmatic/commands/completion.py
@@ -108,25 +108,17 @@ def fish_completion():
         )
         + ('\n# subparser completions',)
         + tuple(
-            '''complete -c borgmatic -a '%s' -d %s -f -n "not __fish_seen_subcommand_from %s"'''
-            % (action_name, shlex.quote(subparser.description), all_subparsers)
+            f'''complete -c borgmatic -a '{action_name}' -d {shlex.quote(subparser.description)} -f -n "not __fish_seen_subcommand_from {all_subparsers}"'''
             for action_name, subparser in subparsers.choices.items()
         )
         + ('\n# global flags',)
         + tuple(
-            '''complete -c borgmatic -a '%s' -d %s %s'''
-            % (' '.join(action.option_strings), shlex.quote(action.help), build_fish_flags(action))
+            f'''complete -c borgmatic -a '{' '.join(action.option_strings)}' -d {shlex.quote(action.help)} {build_fish_flags(action)}'''
             for action in top_level_parser._actions
         )
         + ('\n# subparser flags',)
         + tuple(
-            '''complete -c borgmatic -a '%s' -d %s -n "__fish_seen_subcommand_from %s" %s'''
-            % (
-                ' '.join(action.option_strings),
-                shlex.quote(action.help),
-                action_name,
-                build_fish_flags(action),
-            )
+            f'''complete -c borgmatic -a '{' '.join(action.option_strings)}' -d {shlex.quote(action.help)} -n "__fish_seen_subcommand_from {action_name}" {build_fish_flags(action)}'''
             for action_name, subparser in subparsers.choices.items()
             for action in subparser._actions
         )
--- a/borgmatic/commands/completion.py
+++ b/borgmatic/commands/completion.py
@@ -75,7 +75,11 @@ file_metavars = (
 file_destinations = 'config_paths'
 
 
-def conditionally_emit_arg_completion(action: Action):
+def has_exact_options(action: Action):
+    return action.metavar in file_metavars or action.dest in file_destinations or action.choices
+
+
+def exact_options_completion(action: Action):
     '''
     Given an argparse.Action instance, return a completion invocation
     that forces file completion or options completion, if the action
@@ -84,21 +88,20 @@ def conditionally_emit_arg_completion(action: Action):
     Otherwise, return an empty string.
     '''
 
+    if not has_exact_options(action):
+        return ''
+
     args = ' '.join(action.option_strings)
 
     if action.metavar in file_metavars or action.dest in file_destinations:
-        return dedent(
-            f'''
-        complete -c borgmatic -a '{args}' -Fr -n "__borgmatic_last_arg {args}"'''
-        )
+        return f'''\ncomplete -c borgmatic -Fr -a '{args}' -n "__borgmatic_last_arg {args}"'''
 
     if action.choices:
-        return dedent(
-            f'''
-        complete -c borgmatic -a '{' '.join(map(str, action.choices))}' -n "__borgmatic_last_arg {args}"'''
-        )
+        return f'''\ncomplete -c borgmatic -f -a '{' '.join(map(str, action.choices))}' -n "__borgmatic_last_arg {args}"'''
 
-    return ''
+    raise RuntimeError(
+        f'Unexpected action: {action} passes has_exact_options but has no choices produced'
+    )
 
 
 def dedent_strip_as_tuple(string: str):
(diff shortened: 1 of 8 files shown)


witten (changes requested):
This looks great! Thank you so much for taking this on and updating the docs, too.

The one thing I see as missing is automated testing. Fortunately, the level of testing done for the Bash completions and needed (IMO) for Fish completions is super minimal and high-level. For instance, see `tests/integration/commands/test_completion.py` and `tests/end-to-end/test_completion.py`. They test, respectively, that the completions: 1. Don't raise (integration test), and 2. Actually get interpreted as valid Bash (end-to-end) test.

For adding the latter for Fish, you'll need to edit line 20-21 of `scripts/run-full-tests` and add `fish` to the `apk add` command so that it actually gets installed for use in end-to-end tests.

borgmatic/commands/completion.py
```diff
@@ -55,3 +58,55 @@ def bash_completion():
...
+            for action in top_level_parser._actions
+        ) + (
+            '\n# subparser flags',
+        ) + tuple(
+            '''complete -c borgmatic -a '%s' -d %s -n "__fish_seen_subcommand_from %s" %s'''
+            % (' '.join(action.option_strings), shlex.quote(action.help), actionStr, build_fish_flags(action))
+            for actionStr, subparser in subparsers.choices.items()
+            for action in subparser._actions
+        )
+    )
```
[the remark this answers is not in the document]
isaec: yep! its not quite as good as I would like it to be, but I wasn't able to combine it to group equivalent arguments. Fish doesn't seem to really appreciate the "everything is actually an argument" approach to argparsing, it really wants them to be more like flags associated with args.

However, it has the correct arguments with descriptions, and only shows the contextual args contextually, so that's pretty decent.
scr_13:11:10
image

In particular, knowing when file completions are correct seems to be a bit busted...
image
I can try and fix that, it seems like the conditional arg gets associated in some way.

borgmatic/commands/completion.py
```diff
@@ -55,3 +58,55 @@ def bash_completion():
...
+            '\n# global flags',
+        ) + tuple(
+            '''complete -c borgmatic -a '%s' -d %s %s'''
+            % (' '.join(action.option_strings), shlex.quote(action.help), build_fish_flags(action))
+            for action in top_level_parser._actions
+        ) + (
+            '\n# subparser flags',
+        ) + tuple(
+            '''complete -c borgmatic -a '%s' -d %s -n "__fish_seen_subcommand_from %s" %s'''
+            % (' '.join(action.option_strings), shlex.quote(action.help), actionStr, build_fish_flags(action))
```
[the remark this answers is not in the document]
isaec: yep, that would work, I think I got frustrated with trying to put quotes inside fstrings or escape characters and gave up, but the final iteration should be fine with fstrings. I'll make these changes.

borgmatic/commands/completion.py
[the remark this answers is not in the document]
witten: Pretty cool! Much nicer than the Bash completions, at least the way it's presented to the user. It's up to you whether it's worth sinking more time into the file completions or not.

isaec:
I went a little overboard, and fixed the file completions as well as adding support for the choice style completions, and requiring a value when nargs dictates it. I think I resolved all the feedback on the first draft (fish v Fish, unit and integration tests, fstrings) although I doubtless made some new messes.
image
image
image

The unit tests pass on my system and I can see the new fish integration and unit test in the output, so I figure it's working. It might be interesting to call ``complete -C`` in the integration test to see if the completions are actually being added, but this is likely not a useful test.

Getting the more powerful contextual completions right in a few cases required sorta targeted (and thus brittle) checks, although I think they would degrade gracefully.

I considered archive selection having the possible archives, but I think this is a can of worms on systems like mine where borgmatic can only be called as sudo and completions don't execute as sudo.

I also made the file match check async, which reduces the delay in initial completion display by ~100ms on my system.

Thank you for your wonderful project and your time reviewing my PR!

witten:
Wow, super cool! This now goes way further than the existing Bash completion code, which does file completions but that's pretty much it when it comes to flag values.

> I considered archive selection having the possible archives, but I think this is a can of worms on systems like mine where borgmatic can only be called as sudo and completions don't execute as sudo.

Yeah, that seems potentially thorny to get right.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+
+
+# fish section
+
+
+def has_file_options(action: Action):
+    return action.metavar in (
+        'FILENAME',
+        'PATH',
+    ) or action.dest in ('config_paths',)
```
witten, line 77: Nice! I agree that this is probably fine, especially given that nothing blows up if there's a false negative.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+
+def has_required_param_options(action: Action):
+    return (
+        action.required is True
+        or action.nargs
+        in (
+            '+',
+            '*',
+        )
+        or '--archive' in action.option_strings
```
witten, line 92: I noticed that above you do `action.dest in ('config_paths',)`, but here you check the `option_strings`. However it might be nice to pick one approach and do it everywhere you're trying to identify a particular command-line flag.

Do not feel strongly.

witten: Also, what's the purpose of this particular check? Just to match the couple of actions which have an optional rather than required `--archive` flag? But then isn't that kind of violating the name of the function.. `has_required_param_options()`? (This is definitely a case where a docstring for the function would help me out!)

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+        or action.metavar in ('PATTERN', 'KEYS', 'N')
```
witten, line 93: Similar here. This looks like it's matching optional flags, which I'm assuming _is_ the intent because you went to the trouble of picking them out here.. I just don't know why.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+    return (
+        has_file_options(action) or has_choice_options(action) or has_required_param_options(action)
+    )
+
+
+def exact_options_completion(action: Action):
+    '''
+    Given an argparse.Action instance, return a completion invocation
+    that forces file completion or options completion, if the action
+    takes such an argument and was the last action on the command line.
```
witten, line 108: FYI you can wrap at 100 characters in this codebase.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+    if has_file_options(action):
+        return f'''\ncomplete -c borgmatic -Fr -n "__borgmatic_last_arg {args}"'''
+
+    if has_choice_options(action):
+        return f'''\ncomplete -c borgmatic -f -a '{' '.join(map(str, action.choices))}' -n "__borgmatic_last_arg {args}"'''
+
+    if has_required_param_options(action):
+        return f'''\ncomplete -c borgmatic -x -n "__borgmatic_last_arg {args}"'''
+
+    raise RuntimeError(
```
witten, line 127: `ValueError` might be a little more specific (and still appropriate) here.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+        f'Unexpected action: {action} passes has_exact_options but has no choices produced'
```
witten, line 128: Given that this error message could make its way to a user, it might be a good idea to make it somewhat more user-centric (at the cost of specificity). E.g.: `Could not produce a tab completion for action {action}` or whatever. You can always include more information for the hapless debugging dev with a comment.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+    )
+
+
+def dedent_strip_as_tuple(string: str):
+    return (dedent(string).strip('\n'),)
```
witten, line 133: So here's another thing that might seem like overkill: Even though all of these nicely separated utility functions are covered by the big ol' integration / end-to-end tests you added, IMO they should also have unit test coverage. Part of that is because they _are_ so nicely separated and atomic—you can pretty easily pass in a single value (or a mock) and test that each function returns what it's supposed to.

In cases where a function really has no logic to speak of, a simple "does not raise" unit test would be sufficient IMO.

And I still believe though that you don't need any unit testing on `fish_completion()` itself. The cost/benefit there just wouldn't be worth it.

borgmatic/commands/completion.py
```diff
@@ -55,3 +65,153 @@ def bash_completion():
...
+                            'borgmatic --fish-completion | sudo tee $this_filename',
+                            '$this_filename',
+                        )}"
+                        end
+                    end
+                ' &
+            end
+            __borgmatic_check_version
+
+            function __borgmatic_last_arg --description 'Check if any of the given arguments are the last on the command line'
```
witten, line 179: What if someone is trying to do a completion in the middle of a line?

witten (changes requested):
The overall approach looks good to me. I'm envious about these choice completions and may have to steal that for Bash! I think my main feedback is that I'd love to see unit tests and docstrings for the helpful utility functions you added. I'd be happy to help answer any questions about testing those if you're up to doing it.

borgmatic/commands/completion.py
isaec: image
Great question, works fine, but that means the comment is wrong--should say prior to cursor or something... the specific flags passed to commandline ignore anything after the cursor.

borgmatic/commands/completion.py
isaec: can this error make it to a user? The unit test should fail and as far as I am aware actions are not created on the fly, I think this serves more as an impossible case where you forget to handle one of the cases you filtered for.

witten: Okay, got it.

borgmatic/commands/completion.py
isaec: Is there a way to automate this here? The formatters don't seem to do it and I don't have a good eye for character count.

borgmatic/commands/completion.py
isaec: I think I did this? Not sure if the tests I wrote were quite what you were looking for.

borgmatic/commands/completion.py
isaec: The purpose is to prevent the normal completions from showing up when you hit tab. Given the default is already set for archive, its sorta weird to use it without specifying an archive... but I removed it. Much more helpful if the completions could know about archives. Could borgmatic perhaps have a method that returns a cached list of archives from the last time it interacted with the server, that doesn't require auth?

isaec:
Renamed the PR because these aren't basic anymore :laughing: 
I added unit tests, although they don't look quite like the style you have elsewhere.
I added some doccomments although they might not be clear enough...

borgmatic/commands/completion.py
witten: Depends on your IDE.

 * VSCode: https://stackoverflow.com/questions/43122175/automatically-hard-wrap-lines-at-column-in-vscode
 * vim: https://vi.stackexchange.com/questions/18798/automatic-wraping-to-new-line

borgmatic/commands/completion.py
witten: Thanks for the explanation! A cached list of archives would be possible—but probably not as part of this initial ticket. Feel free to file a new one for that feature though. It might be a little tricky because borgmatic doesn't (currently) list archives upon create. Although I guess it could implicitly append any newly created archive names to the cache without having to run `list`...

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
+from argparse import Action
+from collections import namedtuple
+from typing import Tuple
+
+import pytest
+
+from borgmatic.commands.completion import (
```
witten, line 7: So the convention in other tests in borgmatic would be to do this instead:

```python
from borgmatic.commands import completion as module
```

And then wherever you need to refer to a unit under test, it would look like `module.dedent_strip_as_tuple` or `module.exact_options_completion`.

The intent of this convention, BTW, is that it makes it real easy to identify the unit under test when reading a test function.

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
...
+
+test_data: list[TestCase] = [
+    (Action('--flag', 'flag'), OptionType(file=False, choice=False, unknown_required=False)),
+    *(
+        (
+            Action('--flag', 'flag', metavar=metavar),
+            OptionType(file=True, choice=False, unknown_required=False),
+        )
+        for metavar in ('FILENAME', 'PATH')
+    ),
```
witten, line 27: Clever.

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
...
+from borgmatic.commands.completion import (
+    dedent_strip_as_tuple,
+    exact_options_completion,
+    has_choice_options,
+    has_exact_options,
+    has_file_options,
+    has_unknown_required_param_options,
+)
+
+OptionType = namedtuple('OptionType', ['file', 'choice', 'unknown_required'])
```
witten, line 16: I think this is a prime candidate for a `flexmock()` instance. It _should_ be a drop-in replacement. Although maybe the one thing you can't do easily with a flexmock instance is the `if True in ...` you have below.

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
...
+        OptionType(file=True, choice=False, unknown_required=False),
+    ),
+]
+
+
+@pytest.mark.parametrize('action, option_type', test_data)
+def test_has_file_options_detects_file_options(action: Action, option_type: OptionType):
+    assert (
+        has_file_options(action) == option_type.file
+    ), f'Action: {action} should be file={option_type.file}'
```
witten, line 96: IMO you might not need these messages. The default pytest assert introspection is pretty good these days and will produce a useful error message on its own. (Unless of course you've found otherwise...)

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
...
+
+@pytest.mark.parametrize('action, option_type', test_data)
+def test_has_exact_options_detects_exact_options(action: Action, option_type: OptionType):
+    assert has_exact_options(action) == (
+        True in option_type
+    ), f'Action: {action} should have exact options given {option_type}'
+
+
+@pytest.mark.parametrize('action, option_type', test_data)
+def test_produce_exact_options_completion(action: Action, option_type: OptionType):
```
witten, line 123: Maybe remove the `produce_` here so it's clear what unit this is testing by the function name? You can always put more description after the test name if you need it. E.g. `test_exact_options_completion_produces_blah_blah_blah(...)`.

tests/unit/commands/test_completions.py
```diff
@@ -0,0 +1,143 @@
...
+                '\ncomplete -c borgmatic'
+            ), f'Completion should start with "complete -c borgmatic", got {completion}'
+        else:
+            assert completion == '', f'Completion should be empty, got {completion}'
+
+    except ValueError as value_error:
+        assert False, f'exact_options_completion raised ValueError: {value_error}'
+
+
+def test_dedent_strip_as_tuple():
```
witten, line 137: I would add `_does_not_raise` to the end of this function name to make it abundantly clear what your test is doing.

witten (changes requested):
These tests look great! I really appreciate you taking the time to add them. I left a few minor comments for revision, but I think after that this is basically done. (Assuming you're willing for it to be. :smile:)

tests/unit/commands/test_completions.py
isaec: Why would I use a mock for this? My goal was to create a way to reuse the test cases between the different functions when some inputs would match for multiple functions--I used this tuple to replace having independent lists for each function that were ``(Action, Boolean)``
Is there a reason to mock it out?

tests/unit/commands/test_completions.py
isaec: sounds great, I have no experience with it, I just know assert doesn't produce nice messages in java and reached for an explicit message.

tests/unit/commands/test_completions.py
witten: Nope, I made this comment without realizing that these tuples aren't actually passed in anywhere. Given that they're just for test data structuring, I think they're fine as-is!

tests/unit/commands/test_completions.py
witten: Feel free to play with it and decide for yourself (or not!)...

isaec:
I _hopefully_ didn't create new problems in fixing that, let me know if you have any more feedback.

witten (approved):
Awesome! Thanks again for sticking with this.

witten:
FYI: An unrelated ticket in borgmatic main (https://projects.torsion.org/borgmatic-collective/borgmatic/issues/697) adds support for borgmatic sub-actions, e.g. "borgmatic config bootstrap". This doesn't break the Fish shell completion, but it could work better with these sub-actions. For instance, "bootstrap" should probably not be a completion suggestion unless "config" is specified first, and "list" probably shouldn't be a suggestion after "borgmatic config". I've made these changes to the Bash completion but not to Fish. PRs welcome! :smiley:

isaec:
Thanks for the heads up. If I find myself using sub-actions, I'll take a look at improving things, but for now I think I'll leave this to another motivated user? :grin: It would be good to standardize things but I wonder if some of the specific nuances of when commands "make sense" and how you encode that would resist agnostic utility functions.

The current completions cover everything I use fully.

PR #71, opened 2023-05-15
feat: store configs used to create an archive in the archive and add borgmatic bootstrap

diivi:
From [#697](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/697)

> you need the output of borgmatic/config/collect.py:collect_config_filenames()

🤦 I thought this function will just return the filename and not the complete path (since we use the word "path" in and around these files too).

This should store the config used to create an archive inside the archive itself.

TODO:
- [x] Create a `~/.borgmatic/bootstrap/configs-list.json` file.
- [ ] Update tests

image

### DIFF
--- a/borgmatic/actions/bootstrap.py
+++ b/borgmatic/actions/bootstrap.py
@@ -70,7 +70,7 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
             local_borg_version,
             global_arguments,
             extract_to_stdout=False,
-            destination_path=bootstrap_arguments.destination,
+            destination_path=bootstrap_arguments.destination or '/',
             strip_components=bootstrap_arguments.strip_components,
             progress=bootstrap_arguments.progress,
         )
--- a/borgmatic/actions/bootstrap.py
+++ b/borgmatic/actions/bootstrap.py
@@ -1,0 +1,73 @@
+import logging
+import os
+import json
+
+import borgmatic.borg.extract
+import borgmatic.borg.rlist
+import borgmatic.config.validate
+import borgmatic.hooks.command
+
+from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+
+logger = logging.getLogger(__name__)
+
+def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
+    borgmatic_source_directory = DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    borgmatic_manifest_path = os.path.expanduser(
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
+    )
+    extract_process = borgmatic.borg.extract.extract_archive(
+        global_arguments.dry_run,
+        bootstrap_arguments.repository,
+        borgmatic.borg.rlist.resolve_archive_name(
+            bootstrap_arguments.repository,
+            bootstrap_arguments.archive or 'latest',
+            {},
+            local_borg_version,
+            global_arguments
+        ),
+        [borgmatic_manifest_path],
+        {},
+        {},
+        local_borg_version,
+        global_arguments,
+        extract_to_stdout=True,
+    )
+
+    manifest_data = json.loads(extract_process.stdout.read())
+
+    return manifest_data['config_paths']
+
+    
+
+
+def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
+    '''
+    Run the "bootstrap" action for the given repository.
+    '''
+    manifest_config_paths = get_config_paths(bootstrap_arguments, global_arguments, local_borg_version)
+
+    for config_path in manifest_config_paths:
+        logger.info('Bootstrapping config path %s', config_path)
+
+        borgmatic.borg.extract.extract_archive(
+            global_arguments.dry_run,
+            bootstrap_arguments.repository,
+            borgmatic.borg.rlist.resolve_archive_name(
+                bootstrap_arguments.repository,
+                bootstrap_arguments.archive or 'latest',
+                {},
+                local_borg_version,
+                global_arguments
+            ),
+            [config_path],
+            {},
+            {},
+            local_borg_version,
+            global_arguments,
+            extract_to_stdout=False,
+        )
+
+        
+
+
--- a/borgmatic/actions/bootstrap.py
+++ b/borgmatic/actions/bootstrap.py
@@ -12,7 +12,7 @@ from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
 logger = logging.getLogger(__name__)
 
 def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
-    borgmatic_source_directory = DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    borgmatic_source_directory = bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
     borgmatic_manifest_path = os.path.expanduser(
         os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
     )
@@ -34,7 +34,11 @@ def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
         extract_to_stdout=True,
     )
 
-    manifest_data = json.loads(extract_process.stdout.read())
+    try:
+        manifest_data = json.loads(extract_process.stdout.read())
+    except json.decoder.JSONDecodeError as error:
+        logger.error('Error parsing manifest data: %s', error)
+        raise
 
     return manifest_data['config_paths']
 
@@ -66,6 +70,9 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
             local_borg_version,
             global_arguments,
             extract_to_stdout=False,
+            destination_path=bootstrap_arguments.destination,
+            strip_components=bootstrap_arguments.strip_components,
+            progress=bootstrap_arguments.progress,
         )
 
         
--- a/borgmatic/actions/config/bootstrap.py
+++ b/borgmatic/actions/config/bootstrap.py
@@ -14,14 +14,14 @@ logger = logging.getLogger(__name__)
 def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
     borgmatic_source_directory = bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
     borgmatic_manifest_path = os.path.expanduser(
-        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'manifest.json')
     )
     extract_process = borgmatic.borg.extract.extract_archive(
         global_arguments.dry_run,
         bootstrap_arguments.repository,
         borgmatic.borg.rlist.resolve_archive_name(
             bootstrap_arguments.repository,
-            bootstrap_arguments.archive or 'latest',
+            bootstrap_arguments.archive,
             {},
             local_borg_version,
             global_arguments
@@ -34,11 +34,7 @@ def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
         extract_to_stdout=True,
     )
 
-    try:
-        manifest_data = json.loads(extract_process.stdout.read())
-    except json.decoder.JSONDecodeError as error:
-        logger.error('Error parsing manifest data: %s', error)
-        raise
+    manifest_data = json.loads(extract_process.stdout.read())
 
     return manifest_data['config_paths']
 
@@ -59,7 +55,7 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
             bootstrap_arguments.repository,
             borgmatic.borg.rlist.resolve_archive_name(
                 bootstrap_arguments.repository,
-                bootstrap_arguments.archive or 'latest',
+                bootstrap_arguments.archive,
                 {},
                 local_borg_version,
                 global_arguments
@@ -70,7 +66,7 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
             local_borg_version,
             global_arguments,
             extract_to_stdout=False,
-            destination_path=bootstrap_arguments.destination or '/',
+            destination_path=bootstrap_arguments.destination,
             strip_components=bootstrap_arguments.strip_components,
             progress=bootstrap_arguments.progress,
         )
--- a/borgmatic/actions/config/bootstrap.py
+++ b/borgmatic/actions/config/bootstrap.py
@@ -12,6 +12,16 @@ logger = logging.getLogger(__name__)
 
 
 def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
+    '''
+    Given:
+    The bootstrap arguments, which include the repository and archive name, borgmatic source directory,
+    destination directory, and whether to strip components.
+    The global arguments, which include the dry run flag
+    and the local borg version,
+    Return:
+    The config paths from the manifest.json file in the borgmatic source directory after extracting it from the
+    repository.
+    '''
     borgmatic_source_directory = (
         bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
     )
--- a/borgmatic/actions/config/bootstrap.py
+++ b/borgmatic/actions/config/bootstrap.py
@@ -1,18 +1,20 @@
+import json
 import logging
 import os
-import json
 
 import borgmatic.borg.extract
 import borgmatic.borg.rlist
 import borgmatic.config.validate
 import borgmatic.hooks.command
-
 from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
 
 logger = logging.getLogger(__name__)
 
+
 def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
-    borgmatic_source_directory = bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    borgmatic_source_directory = (
+        bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    )
     borgmatic_manifest_path = os.path.expanduser(
         os.path.join(borgmatic_source_directory, 'bootstrap', 'manifest.json')
     )
@@ -24,7 +26,7 @@ def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
             bootstrap_arguments.archive,
             {},
             local_borg_version,
-            global_arguments
+            global_arguments,
         ),
         [borgmatic_manifest_path],
         {},
@@ -38,14 +40,14 @@ def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
 
     return manifest_data['config_paths']
 
-    
-
 
 def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
     '''
     Run the "bootstrap" action for the given repository.
     '''
-    manifest_config_paths = get_config_paths(bootstrap_arguments, global_arguments, local_borg_version)
+    manifest_config_paths = get_config_paths(
+        bootstrap_arguments, global_arguments, local_borg_version
+    )
 
     for config_path in manifest_config_paths:
         logger.info('Bootstrapping config path %s', config_path)
@@ -58,7 +60,7 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
                 bootstrap_arguments.archive,
                 {},
                 local_borg_version,
-                global_arguments
+                global_arguments,
             ),
             [config_path],
             {},
@@ -70,7 +72,3 @@ def run_bootstrap(bootstrap_arguments, global_arguments, local_borg_version):
             strip_components=bootstrap_arguments.strip_components,
             progress=bootstrap_arguments.progress,
         )
-
-        
-
-
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -2,6 +2,8 @@ import json
 import os
 import logging
 
+import importlib_metadata
+
 import borgmatic.borg.create
 import borgmatic.config.validate
 import borgmatic.hooks.command
@@ -17,19 +19,31 @@ def create_borgmatic_manifest(location, config_paths, dry_run):
     '''
     Create a borgmatic manifest file to store the paths to the configuration files used to create
     the archive.
-    '''    
+    '''
     if dry_run:
         return
-    
-    borgmatic_source_directory = location.get('borgmatic_source_directory') if location.get('borgmatic_source_directory') else DEFAULT_BORGMATIC_SOURCE_DIRECTORY
 
-    borgmatic_manifest_path = os.path.expanduser(os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json'))
+    borgmatic_source_directory = (
+        location.get('borgmatic_source_directory')
+        if location.get('borgmatic_source_directory')
+        else DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    )
+
+    borgmatic_manifest_path = os.path.expanduser(
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
+    )
 
     if not os.path.exists(borgmatic_manifest_path):
         os.makedirs(os.path.dirname(borgmatic_manifest_path), exist_ok=True)
 
     with open(borgmatic_manifest_path, 'w') as f:
-        json.dump(config_paths, f)
+        json.dump(
+            {
+                'borgmatic_version': importlib_metadata.version('borgmatic'),
+                'config_paths': config_paths,
+            },
+            f,
+        )
 
 
 def run_create(
@@ -81,7 +95,9 @@ def run_create(
         location,
         global_arguments.dry_run,
     )
-    create_borgmatic_manifest(location, global_arguments.config_paths, global_arguments.dry_run)
+    create_borgmatic_manifest(
+        location, global_arguments.used_config_paths, global_arguments.dry_run
+    )
     stream_processes = [process for processes in active_dumps.values() for process in processes]
 
     json_output = borgmatic.borg.create.create_archive(
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -1,15 +1,17 @@
 import json
-import os
 import logging
+import os
 
-import importlib_metadata
+try:
+    import importlib_metadata
+except ModuleNotFoundError:  # pragma: nocover
+    import importlib.metadata as importlib_metadata
 
 import borgmatic.borg.create
 import borgmatic.config.validate
 import borgmatic.hooks.command
 import borgmatic.hooks.dispatch
 import borgmatic.hooks.dump
-
 from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
 
 logger = logging.getLogger(__name__)
@@ -23,10 +25,8 @@ def create_borgmatic_manifest(location, config_paths, dry_run):
     if dry_run:
         return
 
-    borgmatic_source_directory = (
-        location.get('borgmatic_source_directory')
-        if location.get('borgmatic_source_directory')
-        else DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    borgmatic_source_directory = location.get(
+        'borgmatic_source_directory', DEFAULT_BORGMATIC_SOURCE_DIRECTORY
     )
 
     borgmatic_manifest_path = os.path.expanduser(
@@ -36,13 +36,13 @@ def create_borgmatic_manifest(location, config_paths, dry_run):
     if not os.path.exists(borgmatic_manifest_path):
         os.makedirs(os.path.dirname(borgmatic_manifest_path), exist_ok=True)
 
-    with open(borgmatic_manifest_path, 'w') as f:
+    with open(borgmatic_manifest_path, 'w') as config_list_file:
         json.dump(
             {
                 'borgmatic_version': importlib_metadata.version('borgmatic'),
                 'config_paths': config_paths,
             },
-            f,
+            config_list_file,
         )
 
 
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -1,4 +1,5 @@
 import json
+import os
 import logging
 
 import borgmatic.borg.create
@@ -7,7 +8,28 @@ import borgmatic.hooks.command
 import borgmatic.hooks.dispatch
 import borgmatic.hooks.dump
 
+from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+
 logger = logging.getLogger(__name__)
+
+
+def create_borgmatic_manifest(location, config_paths, dry_run):
+    '''
+    Create a borgmatic manifest file to store the paths to the configuration files used to create
+    the archive.
+    '''    
+    if dry_run:
+        return
+    
+    borgmatic_source_directory = location.get('borgmatic_source_directory') if location.get('borgmatic_source_directory') else DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+
+    borgmatic_manifest_path = os.path.expanduser(os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json'))
+
+    if not os.path.exists(borgmatic_manifest_path):
+        os.makedirs(os.path.dirname(borgmatic_manifest_path), exist_ok=True)
+
+    with open(borgmatic_manifest_path, 'w') as f:
+        json.dump(config_paths, f)
 
 
 def run_create(
@@ -59,6 +81,7 @@ def run_create(
         location,
         global_arguments.dry_run,
     )
+    create_borgmatic_manifest(location, global_arguments.config_paths, global_arguments.dry_run)
     stream_processes = [process for processes in active_dumps.values() for process in processes]
 
     json_output = borgmatic.borg.create.create_archive(
--- a/borgmatic/actions/create.py
+++ b/borgmatic/actions/create.py
@@ -30,7 +30,7 @@ def create_borgmatic_manifest(location, config_paths, dry_run):
     )
 
     borgmatic_manifest_path = os.path.expanduser(
-        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'manifest.json')
     )
 
     if not os.path.exists(borgmatic_manifest_path):
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -351,7 +351,9 @@ def create_archive(
     sources = deduplicate_directories(
         map_directories_to_devices(
             expand_directories(
-                tuple(location_config.get('source_directories', ())) + borgmatic_source_directories + tuple(global_arguments.config_paths)
+                tuple(location_config.get('source_directories', ()))
+                + borgmatic_source_directories
+                + tuple(global_arguments.used_config_paths)
             )
         ),
         additional_directory_devices=map_directories_to_devices(
--- a/borgmatic/borg/create.py
+++ b/borgmatic/borg/create.py
@@ -351,7 +351,7 @@ def create_archive(
     sources = deduplicate_directories(
         map_directories_to_devices(
             expand_directories(
-                tuple(location_config.get('source_directories', ())) + borgmatic_source_directories
+                tuple(location_config.get('source_directories', ())) + borgmatic_source_directories + tuple(global_arguments.config_paths)
             )
         ),
         additional_directory_devices=map_directories_to_devices(
--- a/borgmatic/commands/arguments.py
+++ b/borgmatic/commands/arguments.py
@@ -9,6 +9,8 @@ SUBPARSER_ALIASES = {
     'compact': [],
     'create': ['-C'],
     'check': ['-k'],
+    'config': [],
+    'config_bootstrap': [],
     'extract': ['-x'],
     'export-tar': [],
     'mount': ['-m'],
@@ -521,6 +523,66 @@ def make_parsers():
     )
     extract_group.add_argument(
         '-h', '--help', action='help', help='Show this help message and exit'
+    )
+
+    config_parser = subparsers.add_parser(
+        'config',
+        aliases=SUBPARSER_ALIASES['config'],
+        help='Perform configuration file related operations',
+        description='Perform configuration file related operations',
+        add_help=False,
+        parents=[top_level_parser],
+    )
+
+    config_subparsers = config_parser.add_subparsers(
+        title='config subcommands',
+        description='Valid subcommands for config',
+        help='Additional help',
+    )
+
+    config_bootstrap_parser = config_subparsers.add_parser(
+        'bootstrap',
+        aliases=SUBPARSER_ALIASES['config_bootstrap'],
+        help='Extract files from a borgmatic created repository to the current directory',
+        description='Extract a named archive from a borgmatic created repository to the current directory without a configuration file',
+        add_help=False,
+        parents=[config_parser],
+    )
+    config_bootstrap_group = config_bootstrap_parser.add_argument_group('config bootstrap arguments')
+    config_bootstrap_group.add_argument(
+        '--repository',
+        help='Path of repository to extract',
+        required=True,
+    )
+    config_bootstrap_group.add_argument(
+        '--archive', help='Name of archive to extract, defaults to "latest"'
+    )
+    config_bootstrap_group.add_argument(
+        '--path',
+        '--restore-path',
+        metavar='PATH',
+        nargs='+',
+        dest='paths',
+        help='Paths to extract from archive, defaults to the entire archive',
+    )
+    config_bootstrap_group.add_argument(
+        '--destination',
+        metavar='PATH',
+        dest='destination',
+        help='Directory to extract files into, defaults to the current directory',
+    )
+    config_bootstrap_group.add_argument(
+        '--strip-components',
+        type=lambda number: number if number == 'all' else int(number),
+        metavar='NUMBER',
+        help='Number of leading path components to remove from each extracted path or "all" to strip all leading path components. Skip paths with fewer elements',
+    )
+    config_bootstrap_group.add_argument(
+        '--progress',
+        dest='progress',
+        default=False,
+        action='store_true',
+        help='Display progress for each file as it is extracted',
     )
 
     export_tar_parser = subparsers.add_parser(
--- a/borgmatic/commands/arguments.py
+++ b/borgmatic/commands/arguments.py
@@ -1,5 +1,6 @@
 import argparse
 import collections
+import itertools
 from argparse import Action, ArgumentParser
 
 from borgmatic.config import collect
@@ -75,14 +76,20 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
                         remaining_arguments.remove(item)
 
         try:
-            arguments[canonical_name] = None if canonical_name in subcommand_parsers_mapping else parsed
+            arguments[canonical_name] = (
+                None if canonical_name in subcommand_parsers_mapping else parsed
+            )
         except UnboundLocalError:
             pass
-    
+
     for argument in arguments:
         if not arguments[argument]:
-            if not any(subcommand in arguments for subcommand in subcommand_parsers_mapping[argument]):
-                raise ValueError(f"Missing subcommand for {argument}. Expected one of {subcommand_parsers_mapping[argument]}")
+            if not any(
+                subcommand in arguments for subcommand in subcommand_parsers_mapping[argument]
+            ):
+                raise ValueError(
+                    f'Missing subcommand for {argument}. Expected one of {subcommand_parsers_mapping[argument]}'
+                )
 
     # If no actions are explicitly requested, assume defaults.
     if not arguments and '--help' not in unparsed_arguments and '-h' not in unparsed_arguments:
@@ -95,14 +102,32 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
 
     # Now ask each subparser, one by one, to greedily consume arguments, from last to first. This
     # allows subparsers to consume arguments before their parent subparsers do.
+    remaining_subparser_arguments = []
+
+    # Now ask each subparser, one by one, to greedily consume arguments, from last to first. This
+    # allows subparsers to consume arguments before their parent subparsers do.
     for subparser_name, subparser in reversed(subparsers.items()):
         if subparser_name not in arguments.keys():
             continue
 
         subparser = subparsers[subparser_name]
-        unused_parsed, remaining_arguments = subparser.parse_known_args(
-            [argument for argument in remaining_arguments if argument != subparser_name]
+        unused_parsed, remaining = subparser.parse_known_args(
+            [argument for argument in unparsed_arguments if argument != subparser_name]
         )
+        remaining_subparser_arguments.append(remaining)
+
+    # Determine the remaining arguments that no subparsers have consumed.
+    if remaining_subparser_arguments:
+        remaining_arguments = [
+            argument
+            for argument in dict.fromkeys(
+                itertools.chain.from_iterable(remaining_subparser_arguments)
+            ).keys()
+            if all(
+                argument in subparser_arguments
+                for subparser_arguments in remaining_subparser_arguments
+            )
+        ]
 
     # Special case: If "borg" is present in the arguments, consume all arguments after (+1) the
     # "borg" action.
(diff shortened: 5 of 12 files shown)


borgmatic/commands/borgmatic.py
```diff
@@ -61,6 +61,7 @@ def run_configuration(config_filename, config, arguments):
         for section_name in ('location', 'storage', 'retention', 'consistency', 'hooks')
     )
     global_arguments = arguments['global']
+    global_arguments.config_paths = used_config_paths
```
diivi, line 64: Will this cause a problem?

witten:
> I thought this function will just return the filename and not the complete path (since we use the word "path" in and around these files too).

Feel free to rename if this is inconsistent!

> Will creating and adding content to this file be okay in create.py itself?

Probably so. One thing I'll mention though is that you *may* want to pull some of the responsibilities that are currently in `borgmatic/borg/create.py` up a level into `borgmatic/actions/create.py`. That's because the former is supposed to be Borg-specific wrapper code (e.g. constructing Borg command-line flags), while the latter is supposed to be borgmatic-specific logic. I say "may" because biting off that shift might be more work than you want to do as part of this. The test changes alone would be a fair amount of work.

So even if you don't refactor, one thing you could do is create/add content your manifest file in `borgmatic/actions/create.py` with the idea that it's borgmatic-specific logic and has nothing to do with Borg.

borgmatic/commands/borgmatic.py
diivi: Okay, I'll work on this next for this PR.

Also, I was thinking of adding the borgmatic version used to create the archive inside this manifest too, so I use the correct config with the correct version. What do you think, and where can I get the current version from? A quick ctrl+f didn't help.

diivi:
Will update the tests after this approach gets reviewed.

borgmatic/actions/create.py
```diff
@@ -1,15 +1,51 @@
...
+    '''
+    Create a borgmatic manifest file to store the paths to the configuration files used to create
+    the archive.
+    '''
+    if dry_run:
+        return
+
+    borgmatic_source_directory = (
+        location.get('borgmatic_source_directory')
+        if location.get('borgmatic_source_directory')
```
witten, line 28: I don't think this `if` is necessary since `.get()` returns None if the key is not found...?

borgmatic/actions/create.py
```diff
@@ -1,15 +1,51 @@
...
+    )
+
+    borgmatic_manifest_path = os.path.expanduser(
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
+    )
+
+    if not os.path.exists(borgmatic_manifest_path):
+        os.makedirs(os.path.dirname(borgmatic_manifest_path), exist_ok=True)
+
+    with open(borgmatic_manifest_path, 'w') as f:
```
witten, line 39: Minor code style nit: No single-letter variables—even in cases like this where it's pretty obvious what it is.

witten (commented):
The approach looks good to me!

diivi:
Got a good code refresher; I'll look at how I can make a new Borgmatic command next (I will refer to the extract command's code).

diivi:
@witten should this me merged independently or should i add the command in this PR too?

witten:
It's entirely up to you. The benefit of merging it independently is it would be a smaller change to merge, and then let you focus on the rest of the feature separately. But then again, sometimes it's nice to be able to work on everything together, especially if you think it's likely you'll need to go back and make substantial changes to earlier work.

borgmatic/commands/arguments.py
```diff
@@ -70,7 +73,18 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
...
                         remaining_arguments.remove(item)
 
-        arguments[canonical_name] = parsed
+        if canonical_name not in subcommand_parsers_mapping:
+            arguments[canonical_name] = parsed
+        else:
+            arguments[canonical_name] = None
+
+    for argument in arguments:
+        if arguments[argument] == None:
```
diivi, line 82: if not arguments[argument]:

diivi:
I tested this and was able to extract the config files, will add tests and refactor code today. Can you check if the general approach is correct?

borgmatic/commands/borgmatic.py
```diff
@@ -616,10 +617,24 @@ def collect_configuration_run_summary_logs(configs, arguments):
...
+        local_borg_version = borg_version.local_borg_version({}, 'borg')
+        borgmatic.actions.bootstrap.run_bootstrap(arguments['bootstrap'], arguments['global'], local_borg_version)
+        yield logging.makeLogRecord(
+            dict(
+                levelno=logging.INFO,
+                levelname='INFO',
+                msg='Bootstrap successful',
+            )
+        )
+        return
```
witten, line 637: Given that `bootstrap` can only really be used by itself, I wonder if it's worth erroring if there are other actions on the command-line. (Probably that wouldn't go here, but in `parse_arguments()` or similar.)

borgmatic/commands/borgmatic.py
```diff
@@ -727,6 +742,7 @@ def main():  # pragma: no cover
         sys.exit(0)
 
     config_filenames = tuple(collect.collect_config_filenames(global_arguments.config_paths))
+    global_arguments.used_config_paths = list(config_filenames)
```
witten, line 745: I feel a little weird about this monkey-patching, but not weird enough to raise an objection over it. I guess we'll see how it goes!

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,73 @@
...
+        borgmatic.borg.rlist.resolve_archive_name(
+            bootstrap_arguments.repository,
+            bootstrap_arguments.archive or 'latest',
+            {},
+            local_borg_version,
+            global_arguments
+        ),
+        [borgmatic_manifest_path],
+        {},
+        {},
```
witten, line 31: Given that config won't be available, there might be some additional command-line flags we'll eventually need for duplicating configuration options. But we can probably add those as user feature requests come in rather than trying to do it in advance.

witten (commented):
Overall approach looks great.. Nice work! I left some comments in the review, but I don't think there's anything major in there.

FYI the more I think about this, the more I'm inclined to stop after config loading and call it a day. That in itself would already be hugely helpful when restoring a system, even if the feature doesn't also do a full file extraction. One thing you could even do is print out a message after bootstrapping succeeds with a `borgmatic extract` command-line the user could copy and paste if they then want to restore a whole system with their bootstrapped configs. Just an idea though.

borgmatic/commands/borgmatic.py
diivi: Should that check be added after we decide what other sub commands should be added to the `borgmatic config` action? `parse_arguments()` seems to check for existence of other keywords, but we don't have any other than bootstrap.

borgmatic/commands/arguments.py
```diff
@@ -865,18 +952,21 @@ def make_parsers():
     )
     borg_group.add_argument('-h', '--help', action='help', help='Show this help message and exit')
 
-    return top_level_parser, subparsers
+    return top_level_parser, subparsers, config_subparsers
```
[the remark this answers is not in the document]
diivi: I was having trouble merging them into one, could you tell me how to merge two Subparser objects? I tried merging their choices but that didn't work.

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,73 @@
...
+        local_borg_version,
+        global_arguments,
+        extract_to_stdout=True,
+    )
+
+    manifest_data = json.loads(extract_process.stdout.read())
+
+    return manifest_data['config_paths']
```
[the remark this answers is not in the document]
diivi: I added error handling for JSONDecodeError around these statements, for other errors, should I wrap the extract_archive() call, or does borgmatic handle that already?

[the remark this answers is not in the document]
witten: [It depends on the error type.](https://projects.torsion.org/borgmatic-collective/borgmatic/src/branch/main/borgmatic/commands/borgmatic.py#L140). So generally you want to catch and then raise an error type that the higher-level borgmatic code can handle, ideally including a user-facing error message in it. Grep for `ValueError` to see some examples.

borgmatic/commands/arguments.py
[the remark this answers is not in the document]
witten: I honestly don't know.. I was imagining you'd do something like your existing `choices` merging trick and then stuff that value into a new instance. If that doesn't work out, a couple of different ideas:

 * Instead of returning `subparsers` and `config_subparsers` separately, return them in a tuple or list. The idea is that it could contain any number of subparsers and the calling code would have to handle that.
 * Or, replace `subparsers` and `config_subparsers` with a single return value of subparser _choices_, which you already know how to merge. Although looking at some of the shell completion code that calls `make_parsers()`, there is one case where it also accesses `subparsers._actions` from that return value. So that would have to be accounted for with this approach.

borgmatic/actions/bootstrap.py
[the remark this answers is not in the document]
witten: Oops, looking at your code changes, that linked `except` might occurs `after` you've already processed the `bootstrap` action and returned`. If that's the case, you might need to catch raised errors yourself (in `collect_configuration_run_summary_logs()`) and then yield a log record appropriately. See other the `before_everything` hooks call in that function for an example.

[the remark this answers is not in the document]
witten: Oops, looking at your code changes, that linked `except` might occurs `after` you've already processed the `bootstrap` action and returned. If that's the case, you might need to catch raised errors yourself (in `collect_configuration_run_summary_logs()`) and then yield a log record appropriately. See other the `before_everything` hooks call in that function for an example.

borgmatic/commands/arguments.py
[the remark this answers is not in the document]
diivi: I think I found a way to make it work with the choices merging:

```python

    merged_subparsers = argparse._SubParsersAction(None, None, metavar=None, dest='merged', parser_class=None)

    for name, subparser in subparsers.choices.items():
        merged_subparsers._name_parser_map[name] = subparser
        subparser._name_parser_map = merged_subparsers._name_parser_map

    for name, subparser in config_subparsers.choices.items():
        merged_subparsers._name_parser_map[name] = subparser
        subparser._name_parser_map = merged_subparsers._name_parser_map

    return top_level_parser, merged_subparsers        


def parse_arguments(*unparsed_arguments):
    '''
    Given command-line arguments with which this script was invoked, parse the arguments and return
    them as a dict mapping from subparser name (or "global") to an argparse.Namespace instance.
    '''
    top_level_parser, subparsers = make_parsers()
```

This works, and I guess it wouldn't harm other functions calling this too?

borgmatic/commands/arguments.py
```diff
@@ -865,18 +952,21 @@ def make_parsers():
...
 
 
 def parse_arguments(*unparsed_arguments):
     '''
     Given command-line arguments with which this script was invoked, parse the arguments and return
     them as a dict mapping from subparser name (or "global") to an argparse.Namespace instance.
     '''
-    top_level_parser, subparsers = make_parsers()
+    top_level_parser, subparsers, config_subparsers = make_parsers()
```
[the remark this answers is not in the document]
diivi: Shared a snippet below.

borgmatic/commands/borgmatic.py
diivi: Ah okay gotcha (I didn't even know you could use multiple actions at once). So there should probably be a list of actions somewhere so I can do something like:

```python
    for action in action_names:
        if 'bootstrap' in arguments and action in arguments['bootstrap']:
            raise ValueError(
                'The bootstrap action cannot be used together with other actions.'
            )
```
?

borgmatic/commands/arguments.py
```diff
@@ -66,7 +73,18 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
...
+        else:
+            arguments[canonical_name] = None
+
+    for argument in arguments:
+        if arguments[argument] == None:
+            for subcommand in subcommand_parsers_mapping[argument]:
+                if subcommand not in arguments:
+                    raise ValueError("Missing subcommand for {}. Expected one of {}".format(
+                        argument, subcommand_parsers_mapping[argument]
+                    ))
```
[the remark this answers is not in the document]
diivi: Oops!

To fix this, first I thought of doing this:

```python
if not any(subcommand in arguments for subcommand in subcommand_parsers_mapping[argument]):
                raise ValueError("Missing subcommand for {}. Expected one of {}".format(
                    argument, subcommand_parsers_mapping[argument]
              ))
```

but this checks if *atleast* one subcommand is present in the arguments, we should check if *exactly* one of them is present right? Is there a better way to do that other than this:

```python
subcommand_count = 0
for subcommand in subcommand_parsers_mapping[argument]:
if subcommand in arguments:
    subcommand_count += 1
if subcommand_count != 1:
raise ValueError("Missing subcommand for {}. Expected one of {}".format(
    argument, subcommand_parsers_mapping[argument]
))
```
?

borgmatic/commands/arguments.py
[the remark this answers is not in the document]
witten: Nice. It *probably* won't harm other callers, but the way to find out is to try it! For instance, with this code in place, you could generate and source the bash completion and make sure that its tab completion still works as expected, especially when tab-completing action names.

borgmatic/commands/borgmatic.py
witten: I think it may be even easier than that. How about:

```python
if 'bootstrap' in arguments.keys() and len(arguments.keys()) > 1:
    raise ValueError(...)
```

This would be before the `'global'` key gets set into `arguments` though.

> I didn't even know you could use multiple actions at once

It's handy for use cases like overriding the default borgmatic actions but just omitting one of them.

borgmatic/commands/arguments.py
```diff
@@ -66,7 +73,18 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
                     if item in subparsers:
                         remaining_arguments.remove(item)
 
-        arguments[canonical_name] = parsed
+        if canonical_name not in subcommand_parsers_mapping:
```
witten, line 76: You might consider removing the `not` and reversing the logic. Or even:

```python
arguments[canonical_name] = None if canonical_name in subcommand_parsers_mapping else parsed
```

Up to you!

borgmatic/commands/borgmatic.py
diivi: Thanks, this makes sense!

diivi:
(Tests remaining)

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,80 @@
+import logging
```
witten, line 1: Not a big deal at all right now, but you might consider moving this file to `borgmatic/actions/config/bootstrap.py`.

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,80 @@
...
+import borgmatic.borg.extract
+import borgmatic.borg.rlist
+import borgmatic.config.validate
+import borgmatic.hooks.command
+
+from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+
+logger = logging.getLogger(__name__)
+
+def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
```
witten, line 14: As you do tests, it would be great if you could also flesh out these functions with docstrings. That includes basic info what arguments it accepts, what it returns, what it _does_, and stuff like "what exception types does this thing raise and when?" (Grep the codebase for "Raise" for examples.)

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,80 @@
...
+    borgmatic_source_directory = bootstrap_arguments.borgmatic_source_directory or DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+    borgmatic_manifest_path = os.path.expanduser(
+        os.path.join(borgmatic_source_directory, 'bootstrap', 'configs-list.json')
```
witten, line 17: You could just call this `manifest.json`. Less specific, but maybe a tiny bit more future-proofed.

borgmatic/actions/bootstrap.py
```diff
@@ -0,0 +1,80 @@
...
+                local_borg_version,
+                global_arguments
+            ),
+            [config_path],
+            {},
+            {},
+            local_borg_version,
+            global_arguments,
+            extract_to_stdout=False,
+            destination_path=bootstrap_arguments.destination or '/',
```
witten, line 73: Maybe make this the default at the arguments level instead of here? Just so it has the "correct" value sooner.

borgmatic/commands/arguments.py
```diff
@@ -66,7 +74,15 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
                     if item in subparsers:
                         remaining_arguments.remove(item)
 
-        arguments[canonical_name] = parsed
+    arguments[canonical_name] = None if canonical_name in subcommand_parsers_mapping else parsed
+
+    for argument in arguments:
+        if not arguments[argument]:
+            if not any(subcommand in arguments for subcommand in subcommand_parsers_mapping[argument]):
+                raise ValueError("Missing subcommand for {}. Expected one of {}".format(
```
witten, line 82: Code style nit: f-string instead of `.format()`.

borgmatic/commands/arguments.py
```diff
@@ -40,6 +43,9 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
         for subparser_name, aliases in SUBPARSER_ALIASES.items()
         for alias in aliases
     }
+    subcommand_parsers_mapping = {
+        'config': ['bootstrap'],
+    }
```
witten, line 48: There might be a way to introspect this information from `subparsers` instead of duplicating it here. (Do not feel strongly.)

borgmatic/commands/arguments.py
```diff
@@ -523,6 +542,68 @@ def make_parsers():
...
+        title='config subcommands',
+        description='Valid subcommands for config',
+        help='Additional help',
+    )
+
+    config_bootstrap_parser = config_subparsers.add_parser(
+        'bootstrap',
+        aliases=SUBPARSER_ALIASES['config_bootstrap'],
+        help='Extract files from a borgmatic created repository to the current directory',
+        description='Extract a named archive from a borgmatic created repository to the current directory without a configuration file',
```
witten, line 568: Note that this help/description are currently inaccurate. You might want to update it to reflect the current code.

borgmatic/commands/arguments.py
```diff
@@ -523,6 +542,68 @@ def make_parsers():
...
+    )
+    config_bootstrap_group = config_bootstrap_parser.add_argument_group('config bootstrap arguments')
+    config_bootstrap_group.add_argument(
+        '--repository',
+        help='Path of repository to extract',
+        required=True,
+    )
+    config_bootstrap_group.add_argument(
+        '--borgmatic-source-directory',
+        help='Path of the borgmatic source directory if other than the default',
```
witten, line 579: Might be nice to explain what this is for users just browsing `--help`. See the `borgmatic_source_directory` description in the config schema for inspiration (or just copying outright).

borgmatic/commands/arguments.py
```diff
@@ -523,6 +542,68 @@ def make_parsers():
...
+        'bootstrap',
+        aliases=SUBPARSER_ALIASES['config_bootstrap'],
+        help='Extract files from a borgmatic created repository to the current directory',
+        description='Extract a named archive from a borgmatic created repository to the current directory without a configuration file',
+        add_help=False,
+    )
+    config_bootstrap_group = config_bootstrap_parser.add_argument_group('config bootstrap arguments')
+    config_bootstrap_group.add_argument(
+        '--repository',
+        help='Path of repository to extract',
```
witten, line 574: Maybe change "extract" to "bootstrap from" or "extract config files from" here.

borgmatic/commands/arguments.py
```diff
@@ -523,6 +542,68 @@ def make_parsers():
...
+        required=True,
+    )
+    config_bootstrap_group.add_argument(
+        '--borgmatic-source-directory',
+        help='Path of the borgmatic source directory if other than the default',
+    )
+    config_bootstrap_group.add_argument(
+        '--archive', help='Name of archive to extract, defaults to "latest"'
```
witten, line 582: Similar here as above: Update "extract" as appropriate.

borgmatic/commands/arguments.py
```diff
@@ -523,6 +542,68 @@ def make_parsers():
...
+    )
+    config_bootstrap_group.add_argument(
+        '--destination',
+        metavar='PATH',
+        dest='destination',
+        help='Directory to extract files into, defaults to /',
```
witten, line 588: And here.

borgmatic/commands/borgmatic.py
```diff
@@ -616,10 +617,27 @@ def collect_configuration_run_summary_logs(configs, arguments):
...
+        try:
+            borgmatic.actions.bootstrap.run_bootstrap(arguments['bootstrap'], arguments['global'], local_borg_version)
+            yield logging.makeLogRecord(
+                dict(
+                    levelno=logging.INFO,
+                    levelname='INFO',
+                    msg='Bootstrap successful',
+                )
+            )
+        except (CalledProcessError, ValueError, OSError) as error:
```
witten, line 638: I don't think this would catch your `JSONDecodeError` or a `KeyError` exception types. So one option is to catch them and re-raise `ValueError`s within `get_config_paths()` or `run_bootstrap()`.. or you can just catch the expected exception types here.

witten (commented):
Looks awesome! I went a little harder on the nitpicking this time since you're about to test, but still nothing major.

diivi:
These tests fail because of the ordering (fixed after I remove the reversed function):

FAILED tests/integration/commands/test_arguments.py::test_parse_arguments_allows_progress_and_extract - SystemExit: 2
FAILED tests/integration/commands/test_arguments.py::test_parse_arguments_extract_with_check_only_extract_does_not_raise - SystemExit: 2
FAILED tests/unit/commands/test_arguments.py::test_parse_subparser_arguments_consumes_multiple_subparser_arguments - AssertionError: assert ['--bar', '3'] == []
FAILED tests/unit/commands/test_arguments.py::test_parse_subparser_arguments_respects_command_line_action_ordering - AssertionError: assert ['--foo', 'true'] == []
FAILED tests/unit/commands/test_arguments.py::test_parse_subparser_arguments_applies_default_subparsers - AssertionError: assert ['--progress'] == []

witten:
Ugh, I was worried about side effects from that `reversed()` "fix" when I suggested it. I'm not sure why reversing the subparsers would cause the `--archive` value not to be seen though. I'm looking into it now to see if I can come up with any suggestions...

witten:
What's happening when `reversed()` is in place is that the `list` subparser is greedily consuming the `--archive` flag such that when it comes time for `extract`'s to consume arguments, that `--archive` flag is already gone. Hence the error. Still trying to figure out some options for what to do about it.

diivi:
Hm, yeah, that's what confused me. Without reversed, shouldn't extract consume it and list won't  have an archive to work with? Oh wait, so I guess list doesn't "require" an archive but extract does.

witten:
Okay, I have a potential fix that allows you to continue using `reversed()`, but you'll probably need to vet it some if you go with this approach. So in that loop where `reversed()` is called, change this:

```python
        unused_parsed, remaining_arguments = subparser.parse_known_args(
            [argument for argument in remaining_arguments if argument != subparser_name]
        )
```

... to this:

```python
        unused_parsed, remaining_arguments = subparser.parse_known_args(
            [argument for argument in unparsed_arguments if argument != subparser_name]
        )
```

In other words, change `remaining_arguments` to `unparsed_arguments` in this one spot. The idea is that `remaining_arguments` is returned from the `parse_subparser_arguments()` function because the caller needs to pass those `remaining_arguments` to the top-level parser. (If you give that parser _all_ the arguments, it errors.) So we still need to update `remaining_arguments` as each subparser consumes arguments.

But _within_ this function, we actually want to share arguments among subparsers—like sharing `--archive` between `list` and `extract` to prevent the exact error you're seeing.

Let me know if that makes sense and/or if you spot any issues with the approach.

diivi:
Oh, nice, I think I understand, and:
```python
    module.parse_arguments('--progress', 'extract', '--archive', 'test', 'list')
```
passes now. But 
```python
    module.parse_arguments('create', '--progress', 'list')
```
still doesn't (including the other 3 fails I listed).

Shouldn't the fix you mentioned be universal? It just worked for one case 🤔. I did some trial and error but got so confused that I forgot the problem I was solving, your explanation did help though 😅

witten:
Ah, I just tried your command-line example above. I didn't try the tests. My bad! Okay, I see now that with my "fix," `remaining_arguments` isn't getting updated by each successive `parse_known_arguments()` call in that loop. It's only getting updated by the last one. Let me see if I can do anything about that...

> I did some trial and error but got so confused that I forgot the problem I was solving, your explanation did help though

Yeah, this section of the code is unfortunately pretty complex and tough to reason about. I'd love to simplify it, but it is dealing with some pretty complex requirements. Fortunately, it's pretty well tested. So (as you discovered) if you break some implicit functionality, the tests will complain loudly.

witten:
Well, I'm not super happy with this due to the complexity, but here's some code that both works with your example command and passes the tests. If you like, you can use it to replace your existing loop that does the `reversed()` call:

```python
    remaining_subparser_arguments = []

    # Now ask each subparser, one by one, to greedily consume arguments, from last to first. This
    # allows subparsers to consume arguments before their parent subparsers do.
    for subparser_name, subparser in reversed(subparsers.items()):
        if subparser_name not in arguments.keys():
            continue
    
        subparser = subparsers[subparser_name]
        unused_parsed, remaining = subparser.parse_known_args(
            [argument for argument in unparsed_arguments if argument != subparser_name]
        )   
        remaining_subparser_arguments.append(remaining)
        
    if remaining_subparser_arguments:
        remaining_arguments = [
            argument
            for argument in dict.fromkeys(
                itertools.chain.from_iterable(remaining_subparser_arguments)
            ).keys()
            if all(
                argument in subparser_arguments
                for subparser_arguments in remaining_subparser_arguments
            )
        ]
```

Here's my attempt at an explanation. In the first loop, each subparser is asked to parse _all_ the arguments, so each subparser gets the same chance to parse `--archive latest`, for instance. Then, each subparser's remaining arguments are added to a list of lists (`remaining_subparser_arguments`).

Next, in the newly added code, all the remaining arguments from all subparsers get contatenated and uniqued into a single list (`dict.fromkeys(itertools.chain.from_iterable(...))`) and iterated over. That way we're looking at each remaining argument one at a time. We keep each one and put it in `remaining_arguments`—but only if exists in each of the subparser-specific remaining arguments lists.

Therefore, we end up with only the arguments that _all_ the subparsers passed on (didn't parse). Any "remaining" arguments that _some_ subparsers _did_ parse get excluded.

To make this a little more concrete, here's a contained example of that code in use:

```python
remaining_subparser_arguments = (
    ('--latest', 'archive', 'prune', 'extract', 'list', '--wtf'),
    ('--latest', 'archive', 'check', 'extract', 'list', '--wtf'),
    ('prune', 'check', 'list', '--wtf'),
    ('prune', 'check', 'extract', '--wtf'),
)

remaining_arguments = [
    argument
    for argument in dict.fromkeys(itertools.chain.from_iterable(remaining_subparser_arguments)).keys()
    if all(argument in subparser_arguments for subparser_arguments in remaining_subparser_arguments)
]   

print(remaining_arguments)
```

And the result is:

```python
['--wtf']
```

The one thing I'm not entirely sure about the unique-ifying done with the `dict.fromkeys()` call. It's necessary so you don't end up with a duplicated result like `['--wtf', '--wtf', '--wtf', '--wtf']`, but I haven't thought through whether that would break other stuff. ("Duplicated" arguments like `extract` in `borgmatic --progress check --only extract extract --archive latest` does not break, because they don't end up in `remaining_subparser_arguments`, as they're consumed by the subparsers.)

If you don't feel comfortable adding this sort of complexity, please don't feel beholden to this solution. It's just the best I could come up with after banging my head on this for a bit.

diivi:
Damn, nice. I don't really have any idea about making this simpler 😓, maybe I'll have a look at refactoring it in the future again.

I'll use this for now and add tests to cover my own changes, thank you so much!

diivi:
You can review the tests now (last 2 commits), I'll add docstrings later tonight.

borgmatic/actions/create.py
```diff
@@ -59,6 +95,9 @@ def run_create(
         location,
         global_arguments.dry_run,
     )
+    create_borgmatic_manifest(
+        location, global_arguments.used_config_paths, global_arguments.dry_run
+    )
```
witten, line 100: Probably needs mocking in existing `run_create()` unit tests?

borgmatic/borg/create.py
```diff
@@ -351,7 +351,9 @@ def create_archive(
     sources = deduplicate_directories(
         map_directories_to_devices(
             expand_directories(
-                tuple(location_config.get('source_directories', ())) + borgmatic_source_directories
+                tuple(location_config.get('source_directories', ()))
+                + borgmatic_source_directories
+                + tuple(global_arguments.used_config_paths)
```
witten, line 356: Does this need a test? Maybe something along the lines of the existing `test_create_archive_with_patterns_calls_borg_with_patterns_including_converted_source_directories()`?

borgmatic/commands/arguments.py
```diff
@@ -77,13 +95,34 @@ def parse_subparser_arguments(unparsed_arguments, subparsers):
...
+        remaining_arguments = [
+            argument
+            for argument in dict.fromkeys(
+                itertools.chain.from_iterable(remaining_subparser_arguments)
+            ).keys()
+            if all(
+                argument in subparser_arguments
+                for subparser_arguments in remaining_subparser_arguments
+            )
+        ]
```
witten, line 125: IMO this would benefit from some explicit test coverage. I think the easiest way to do that would be to factor it out into a separate function, mock that function within tests of this `parse_subparser_arguments()`, and then independently test the function by passing in various `remaining_subparser_arguments` and asserting the result.

But if you'd rather keep it all inline, you could probably still test the logic by calling `parse_subparser_arguments()` and manipulating a mocked `parse_known_args()` to return the results you want to test.

Let me know though if either approach looks like it will be problematic.

borgmatic/commands/arguments.py
```diff
@@ -865,7 +969,19 @@ def make_parsers():
...
+
+    for name, subparser in subparsers.choices.items():
+        merged_subparsers._name_parser_map[name] = subparser
+        subparser._name_parser_map = merged_subparsers._name_parser_map
+
+    for name, subparser in config_subparsers.choices.items():
+        merged_subparsers._name_parser_map[name] = subparser
+        subparser._name_parser_map = merged_subparsers._name_parser_map
+
+    return top_level_parser, merged_subparsers
```
witten, line 984: Any test for these two loops portion of the function? I'm still not 100% sure of what they do. :smile:

witten (commented):
You've got some nice tests here! Most of my comments are either: 1. Nitpicking on what's mocked or not mocked in the tests, and 2. Requests for additional tests. So nothing too earth-shattering, and hopefully pretty straightforward to follow up on. If you find that's not the case and/or you'd like help testing, please do let me know.

And if it seems like I'm being overly hardcore about including tests about absolutely everything, it's because with backup software, user expectations are pretty high about things not breaking. I'll admit that even so, sometimes I ship totally broken stuff anyway. But my hope is that by being attentive to testing, that will at least catch some of the potential issues or regressions. It *is* often annoying though, and sometimes even I get tired of writing tests! (So please push back if you disagree with any of my opinions on the PR.)

tests/unit/actions/config/test_bootstrap.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+    flexmock(module.borgmatic.borg.extract).should_receive('extract_archive').and_return(
+        extract_process
+    )
+    flexmock(module.borgmatic.borg.rlist).should_receive('resolve_archive_name').and_return(
+        'archive'
+    )
+    flexmock(module.borgmatic.borg.extract).should_receive('extract_archive').and_return(
+        extract_process
+    )
```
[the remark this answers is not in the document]
witten: Sure! If this is a `_does_not_raise()` test, then you don't need to assert anything! Just put the `extract_archive` mock in place once, and it'll be used each time it's called. But instead if you do actually want to assert that `extract_archive()` gets called twice, then the `.twice()` invocation I pasted above should do it.

tests/unit/actions/test_create.py
```diff
@@ -103,3 +103,20 @@ def test_run_create_bails_if_repository_does_not_match():
...
+
+def test_create_borgmatic_manifest_creates_manifest_file():
+    flexmock(module.os.path).should_receive('expanduser').and_return('/home/user')
+    flexmock(module.os.path).should_receive('join').and_return('/home/user/bootstrap/manifest.json')
+    flexmock(module.os.path).should_receive('exists').and_return(False)
+    flexmock(module.os).should_receive('makedirs').and_return(True)
+
+    flexmock(module.json).should_receive('dump').and_return(True)
+
+    module.create_borgmatic_manifest({}, 'test.yaml', False)
```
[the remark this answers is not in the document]
diivi: What difference would that make test-wise? Only this line is concerned with the source directory right - 

```python
    borgmatic_source_directory = location.get(
        'borgmatic_source_directory', DEFAULT_BORGMATIC_SOURCE_DIRECTORY
    )
```

So should I check if I do give it a location, it should use that instead of the DEFAULT directory? I though `.get()` won't induce any unexpected behaviour so I didn't add it.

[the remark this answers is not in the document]
witten: Yeah, it's less about testing that `.get()` does what it's supposed to, but more about testing that the function as a whole (regardless of its implementation) adheres to its contract of honoring the `borgmatic_source_directory` option. If someone comes along later and accidentally deletes that `.get()` line or changes it to hard-code `DEFAULT_BORGMATIC_SOURCE_DIRECTORY`, you ideally want a test failing loudly (and clearly). Getting on my [soapbox](https://idioms.thefreedictionary.com/soapbox) for a moment: Tests should be about defining and enforcing the contract that a unit under test has with its caller. If a part of the contract _isn't_ defined in the tests, then it's not part of the contract IMO!

Having said all that, I don't actually feel strongly about this particular example, so follow your heart. :smile:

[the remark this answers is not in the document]
diivi: > Tests should be about defining and enforcing the contract that a unit under test has with its caller. If a part of the contract isn't defined in the tests, then it's not part of the contract.

True, will remember this quote when writing tests in the future 😄

tests/unit/actions/config/test_bootstrap.py
```diff
@@ -0,0 +1,33 @@
...
+    global_arguments = flexmock(
+        dry_run=False,
+    )
+    local_borg_version = flexmock()
+    extract_process = flexmock(
+        stdout=flexmock(
+            read=lambda: '{"config_paths": ["/borgmatic/config.yaml"]}',
+        ),
+    )
+    flexmock(module.borgmatic.borg.extract).should_receive('extract_archive').and_return(
```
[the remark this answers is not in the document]
witten: One thing though is just to make sure any side-effect-having functions of `get_config_paths()` are mocked out as well in these tests.

borgmatic/commands/arguments.py
diivi: They merge all the subsubparsers and our subparser into one. Here's the discussion - https://github.com/borgmatic-collective/borgmatic/pull/71#discussion_r1212166616

borgmatic/commands/arguments.py
diivi: Shouldn't this be considered tested, considering failing tests prompted us to rewrite this section ( the extract/list thing)? What should I test here?

borgmatic/commands/arguments.py
diivi: Oh, I can't really say I do either, I was just trying things out and this ended up working.

diivi: Now that I look at it again, this line in both loops confuses me too:

```python
        subparser._name_parser_map = merged_subparsers._name_parser_map
```

borgmatic/commands/arguments.py
witten: You're right that it's got _coverage_, meaning the code is exercised by tests, but it's not tested on its own. The code is fairly complex and while it definitely handles the `extract/list` case as you point out, without targeted tests I wouldn't necessarily be convinced it'd handle other cases of finding unique remaining arguments across all subparsers. If this were my feature (and to be clear, it's most definitely not!), I'd probably write a targeted test that, given some contrived `remaining_subparser_arguments`, assert that the code produces the expected `remaining_arguments`. And I might even throw a couple of different variants at it to make sure it handles edge cases like zero overlapping remaining arguments across the subparsers, an empty `remaining_subparser_arguments` element for one subparser, etc etc. Basically whenever I'm scared by complexity I throw test cases at the problem until I'm a little calmer.

Having said all that, this is _your_ feature, so if you're comfortable with the existing coverage, I can be fine with it too.

diivi: Yeah, I guess it could be tested more.

I tried this - [`6475345` (#71)](https://github.com/borgmatic-collective/borgmatic/pull/71/commits/6475345a8fbaa9b3d9aaeb7444075829c49343ab#diff-d39d71411512f243818b758012a11063b00ebfa284f762c22eca562d30d0bce8R180) to parameterize and separate the unit to test it.

But it's probably wrong since whatever I pass as the array here:

borgmatic/commands/arguments.py
```diff
@@ -28,6 +28,37 @@
 }
 
 
+def parse_subparser_arguments(arguments, unparsed_arguments, subparsers):
```
diivi, line 31: parse_subsubparser_arguments?

borgmatic/commands/arguments.py
diivi: Another case that I tried but don't know if this is correct (it passes the test):
```python
        (
            {'extract': flexmock(), 'list': flexmock()},
            ['--progress', '--archive', 'myarchive'],
            {
                'extract': flexmock(parse_known_args=lambda arguments: (flexmock(), ['--archive', 'myarchive'])),
                'list': flexmock(parse_known_args=lambda arguments: (flexmock(), [])),
            },
            [],
        )
```

witten: Nevermind, I think I found the code in question. I'll have a look.

borgmatic/commands/arguments.py
witten: It's not only doing subsubparsers.. This is also the code doing the actual parsing for each top-level subparser (and returning remaining arguments, of course). So maybe `parse_known_subparses_arguments()` or something?

borgmatic/commands/arguments.py
witten: > From https://github.com/borgmatic-collective/borgmatic/pull/71#issuecomment-1577651275, here's some potential test case data:

Note that my example test case data from above is only for the second half of your factored-out `parse_subparser_arguments()`/`parse_known_subparser_arguments` function. (I hadn't found the actual function before I commented.) So if you wanted a test case for the whole function, it would be like one of your existing examples except `unparsed_arguments` might be `('--latest', 'archive', 'check', 'prune', 'extract', 'list', '--wtf')` with the subparsers set up to mimic the actual `check`, `prune`, `extract`, `list`, etc. But maybe your existing `extract`/`list`-only example would be sufficient.

witten: > Another case that I tried but don't know if this is correct (it passes the test):

Okay, finally viewing this test data in light of the actual code. My thoughts:

 * Do you need the action names in the unparsed arguments?
 * Does the `extract` subparser need to consume `--archive` like `list` does?
 * Should the `list` subparser return `--progress` among their remaining arguments? AFAIK it doesn't touch `--progress`.

If you want to "cheat" here, just instrument the actual code with print statements and then run borgmatic with the flags of your test case. Then look at the printed values and steal them for your tests.

borgmatic/commands/arguments.py
witten: Another thought: Sometimes, if a function is tough to name, it means you haven't captured a coherent, describable unit of functionality with it. Like maybe it either does too much or too little such that it's hard to describe succinctly. So one thing you _could_ do in this case is to only make the function contain the "# Determine the remaining arguments that no subparsers have consumed" code and not the other loop. That way it'd be easier to create test cases for and it'd be easier to name. Just an idea.

borgmatic/commands/arguments.py
```diff
@@ -523,6 +569,71 @@ def make_parsers():
...
+        title='config subcommands',
+        description='Valid subcommands for config',
+        help='Additional help',
+    )
+
+    config_bootstrap_parser = config_subparsers.add_parser(
+        'bootstrap',
+        aliases=SUBPARSER_ALIASES['config_bootstrap'],
+        help='Extract the config files used to create a borgmatic repository',
+        description='Extract just the config files that were used to create a borgmatic repository during the "create" operation',
```
witten, line 593: Convention (in borgmatic) is for subparsers to have identical `help` and `description`.

borgmatic/commands/arguments.py
```diff
@@ -523,6 +569,71 @@ def make_parsers():
...
+        '--progress',
+        dest='progress',
+        default=False,
+        action='store_true',
+        help='Display progress for each file as it is extracted',
+    )
+    config_bootstrap_group.add_argument(
+        '-h', '--help', action='help', help='Show this help message and exit'
+    )
+
```
witten, line 636: Something that occurs to me: Did you have a chance to play with [borgmatic's bash completion](https://torsion.org/borgmatic/docs/how-to/set-up-backups/#bash) with these argument changes incorporated? For instance: `borgmatic conf<tab>` or `borgmatic config boot<tab>` or even `borgmatic config bootstrap --prog<tab>`. Might also be good to try just `borgmatic boot<tab>` and see what happens.

If some of these don't work as expected, it'd be good to know. They don't necessarily need to be fixed in this PR (the scope is already large enough), but they should be filed for follow-on work. I might even be able to pick them up after they're filed (if they need to be).

borgmatic/commands/arguments.py
```diff
@@ -24,6 +28,25 @@
 }
 
 
+def get_unparsable_arguments(remaining_subparser_arguments):
+    # Determine the remaining arguments that no subparsers have consumed.
```
witten, line 32: Should be a doc string at this point, IMO.

borgmatic/actions/config/bootstrap.py
```diff
@@ -0,0 +1,74 @@
...
+import borgmatic.borg.extract
+import borgmatic.borg.rlist
+import borgmatic.config.validate
+import borgmatic.hooks.command
+from borgmatic.borg.state import DEFAULT_BORGMATIC_SOURCE_DIRECTORY
+
+logger = logging.getLogger(__name__)
+
+
+def get_config_paths(bootstrap_arguments, global_arguments, local_borg_version):
```
witten, line 14: Docstring missing. (It'd be nice to know what the arguments are, too!)

witten (commented):
This is looking great. We're almost there.. I can almost smell the merge! I noted a few minor things in the tests and a few places in the code. As per usual, let me know if you have any questions.

borgmatic/commands/arguments.py
diivi: Just added " during the "create" operation" now.

borgmatic/commands/arguments.py
diivi: I use zsh, will completions work for that? I remember trying it when you mentioned this, but no completions worked for me (even with older versions). Let me try with bash.

diivi: image
Works!

image

It recommended borg as a subcommand too though

witten: Awesome! Yeah, completions are only implemented for bash and fish right now. zsh would be a logical next step.

Recommending `borg` after `config` should probably be fixed (not in this PR!), as that would produce an invalid command-line. Feel free to file it as a separate issue.

tests/unit/borg/test_create.py
```diff
@@ -582,7 +582,52 @@ def test_create_archive_with_patterns_calls_borg_with_patterns_including_convert
...
+    module.create_archive(
+        dry_run=False,
+        repository_path='repo',
+        location_config={
+            'source_directories': ['foo', 'bar'],
+            'repositories': ['repo'],
+        },
+        storage_config={},
+        local_borg_version='1.2.3',
+        global_arguments=flexmock(log_json=False, used_config_paths=['/etc/borgmatic/config.yaml']),
```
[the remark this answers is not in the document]
diivi: So I did this - 

```python
def test_create_archive_with_sources_and_used_config_paths_calls_borg_with_sources_and_config_paths():
    flexmock(module.borgmatic.logger).should_receive('add_custom_log_levels')
    flexmock(module.logging).ANSWER = module.borgmatic.logger.ANSWER
    flexmock(module).should_receive('collect_borgmatic_source_directories').and_return(['/etc/borgmatic/config.yaml'])
    flexmock(module).should_receive('deduplicate_directories').and_return(
        ('foo', 'bar', '/etc/borgmatic/config.yaml')
    )
    flexmock(module).should_receive('map_directories_to_devices').and_return({})
    flexmock(module).should_receive('expand_directories').with_args(
        ('foo', 'bar', '/etc/borgmatic/config.yaml')
    ).and_return(())
```

But got this error:

"Arguments for call expand_directories did not match expectations:\n  Received call:\texpand_directories(['/etc/borgmatic/config.yaml'])\n  Expected call[1]:\texpand_directories(directories=('foo', 'bar', '/etc/borgmatic/config.yaml'))"

Why didn't expand_directories recieve the location_config as args? And this is what I should be testing right?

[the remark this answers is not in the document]
witten: I'm guessing because `expand_directories()` is called twice—once inside of `map_directories_to_devices()` and once above, on line 346. And I bet that first call is the one where the expectation isn't matched...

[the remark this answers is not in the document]
witten: > And this is what I should be testing right?

And yeah, I think this is probably the right thing to test given the constraints. It's obnoxious to have to do so much mocking, but I don't see a better way to do it short of switching to an integration test.

borgmatic/commands/arguments.py
```diff
@@ -865,7 +978,28 @@ def make_parsers():
     )
     borg_group.add_argument('-h', '--help', action='help', help='Show this help message and exit')
 
-    return top_level_parser, subparsers
+    merged_subparsers = argparse._SubParsersAction(
+        None, None, metavar=None, dest='merged', parser_class=None
+    )
```
witten, line 983: Can be deleted now that it's been factored out?

tests/unit/actions/test_create.py
```diff
@@ -103,3 +108,42 @@ def test_run_create_bails_if_repository_does_not_match():
             remote_path=None,
         )
     )
+
+
+def test_create_borgmatic_manifest_creates_manifest_file():
```
witten, line 113: Since this function doesn't have any asserts (implicit or explict), I'm not sure the name of the function is accurate right now. As written, this function is more of a "does not raise" test case.

tests/unit/actions/test_create.py
```diff
@@ -103,3 +108,42 @@ def test_run_create_bails_if_repository_does_not_match():
...
+    flexmock(module.os.path).should_receive('exists').and_return(False)
+    flexmock(module.os).should_receive('makedirs').and_return(True)
+
+    flexmock(module.importlib_metadata).should_receive('version').and_return('1.0.0')
+    flexmock(module.json).should_receive('dump').and_return(True)
+
+    module.create_borgmatic_manifest({}, 'test.yaml', False)
+
+
+def test_create_borgmatic_manifest_creates_manifest_file_with_custom_borgmatic_source_directory():
```
witten, line 123: Same here.

tests/unit/actions/test_create.py
```diff
@@ -103,3 +108,42 @@ def test_run_create_bails_if_repository_does_not_match():
...
+    )
+    flexmock(module.json).should_receive('dump').and_return(True)
+
+    module.create_borgmatic_manifest(
+        {'borgmatic_source_directory': '/borgmatic'}, 'test.yaml', False
+    )
+
+
+def test_create_borgmatic_manifest_does_not_create_manifest_file_on_dry_run():
+    flexmock(module.os.path).should_receive('expanduser').never()
```
witten, line 147: This one looks good because it has an actual assertion (`.never()`)!

tests/unit/borg/test_create.py
```diff
@@ -582,7 +582,56 @@ def test_create_archive_with_patterns_calls_borg_with_patterns_including_convert
...
+    flexmock(module).should_receive('collect_borgmatic_source_directories').and_return([])
+    flexmock(module).should_receive('deduplicate_directories').and_return(
+        ('foo', 'bar', '/etc/borgmatic/config.yaml')
+    )
+    flexmock(module).should_receive('map_directories_to_devices').and_return({})
+    flexmock(module).should_receive('expand_directories').with_args([]).and_return(())
+    flexmock(module).should_receive('expand_directories').with_args(
+        ('foo', 'bar', '/etc/borgmatic/config.yaml')
+    ).and_return(('foo', 'bar', '/etc/borgmatic/config.yaml'))
+    flexmock(module).should_receive('expand_directories').with_args([]).and_return(())
```
witten, line 601: Thanks for doing this. I know it was a pain...

borgmatic/commands/arguments.py
witten: (I'll do this to save the PR roundtrip.)

tests/unit/actions/test_create.py
witten: (I'll fix this to save the PR roundtrip. Probably just add a `.once()`...)

witten (approved):
I'm calling this done! Thanks so much for slogging all the way to the finish line on this one.. I know it wasn't always easy, and I know in places things got more complicated than you envisioned going in. But I think the end result is a huge achievement—and will be a really nice feature for users to have!

diivi:
Thanks! I learnt a lot of new argparse stuff and working with you has always been a delight!

PR #73, opened 2023-06-12
feat: allow restoring to different port/host/username

diivi:
This PR adds a new feature that allows users to restore to a different port, host, or username. 
This Draft is to show the configuration version for PostgreSQL, I haven't tested this yet, just created to share progress.
Next Steps:

- [x] Provide overrides from the command line.
- [x] Do the same thing for :
- [x] MongoDB
- [x] MySQL
- [x] SQLite.
- [x] Tests

### DIFF
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -68,13 +68,19 @@ def restore_single_database(
     archive_name,
     hook_name,
     database,
+    connection_params,
 ):  # pragma: no cover
     '''
-    Given (among other things) an archive name, a database hook name, and a configured database
+    Given (among other things) an archive name, a database hook name, the hostname,
+    port, username and password as connection params,  and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(
         f'{repository.get("label", repository["path"])}: Restoring database {database["name"]}'
+    )
+
+    logger.info(
+        f'hostname port username password for database {database["name"]}'
     )
 
     dump_pattern = borgmatic.hooks.dispatch.call_hooks(
@@ -113,6 +119,7 @@ def restore_single_database(
         location,
         global_arguments.dry_run,
         extract_process,
+        connection_params,
     )
 
 
@@ -308,6 +315,13 @@ def run_restore(
                 hooks, archive_database_names, hook_name, database_name
             )
 
+            connection_params = {
+                'hostname': restore_arguments.hostname,
+                'port': restore_arguments.port,
+                'username': restore_arguments.username,
+                'password': restore_arguments.password,
+            }
+
             if not found_database:
                 remaining_restore_names.setdefault(found_hook_name or hook_name, []).append(
                     database_name
@@ -327,6 +341,7 @@ def run_restore(
                 archive_name,
                 found_hook_name or hook_name,
                 dict(found_database, **{'schemas': restore_arguments.schemas}),
+                connection_params,
             )
 
     # For any database that weren't found via exact matches in the hooks configuration, try to
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -316,6 +316,7 @@ def run_restore(
                 'port': restore_arguments.port,
                 'username': restore_arguments.username,
                 'password': restore_arguments.password,
+                'restore_path': restore_arguments.restore_path,
             }
 
             if not found_database:
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -304,20 +304,19 @@ def run_restore(
     restore_names = find_databases_to_restore(restore_arguments.databases, archive_database_names)
     found_names = set()
     remaining_restore_names = {}
+    connection_params = {
+        'hostname': restore_arguments.hostname,
+        'port': restore_arguments.port,
+        'username': restore_arguments.username,
+        'password': restore_arguments.password,
+        'restore_path': restore_arguments.restore_path,
+    }
 
     for hook_name, database_names in restore_names.items():
         for database_name in database_names:
             found_hook_name, found_database = get_configured_database(
                 hooks, archive_database_names, hook_name, database_name
             )
-
-            connection_params = {
-                'hostname': restore_arguments.hostname,
-                'port': restore_arguments.port,
-                'username': restore_arguments.username,
-                'password': restore_arguments.password,
-                'restore_path': restore_arguments.restore_path,
-            }
 
             if not found_database:
                 remaining_restore_names.setdefault(found_hook_name or hook_name, []).append(
@@ -368,6 +367,7 @@ def run_restore(
                 archive_name,
                 found_hook_name or hook_name,
                 dict(database, **{'schemas': restore_arguments.schemas}),
+                connection_params,
             )
 
     borgmatic.hooks.dispatch.call_hooks_even_if_unconfigured(
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -79,10 +79,6 @@ def restore_single_database(
         f'{repository.get("label", repository["path"])}: Restoring database {database["name"]}'
     )
 
-    logger.info(
-        f'hostname port username password for database {database["name"]}'
-    )
-
     dump_pattern = borgmatic.hooks.dispatch.call_hooks(
         'make_database_dump_pattern',
         hooks,
--- a/borgmatic/actions/restore.py
+++ b/borgmatic/actions/restore.py
@@ -72,7 +72,7 @@ def restore_single_database(
 ):  # pragma: no cover
     '''
     Given (among other things) an archive name, a database hook name, the hostname,
-    port, username and password as connection params,  and a configured database
+    port, username and password as connection params, and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(
--- a/borgmatic/hooks/mongodb.py
+++ b/borgmatic/hooks/mongodb.py
@@ -143,7 +143,7 @@ def build_restore_command(extract_process, database, dump_filename, connection_p
     Return the mongorestore command from a single database configuration.
     '''
     hostname = connection_params['hostname'] or database.get('restore_hostname', database.get('hostname'))
-    port = str(connection_params['port'] or database.get('restore_port', database.get('port')))
+    port = str(connection_params['port'] or database.get('restore_port', database.get('port', '')))
     username = connection_params['username'] or database.get('restore_username', database.get('username'))
     password = connection_params['password'] or database.get('restore_password', database.get('password'))
 
--- a/borgmatic/hooks/mongodb.py
+++ b/borgmatic/hooks/mongodb.py
@@ -102,7 +102,9 @@ def make_database_dump_pattern(
     return dump.make_database_dump_filename(make_dump_path(location_config), name, hostname='*')
 
 
-def restore_database_dump(database_config, log_prefix, location_config, dry_run, extract_process, connection_params):
+def restore_database_dump(
+    database_config, log_prefix, location_config, dry_run, extract_process, connection_params
+):
     '''
     Restore the given MongoDB database from an extract stream. The database is supplied as a
     one-element sequence containing a dict describing the database, as per the configuration schema.
@@ -122,7 +124,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     dump_filename = dump.make_database_dump_filename(
         make_dump_path(location_config), database['name'], database.get('hostname')
     )
-    restore_command = build_restore_command(extract_process, database, dump_filename, connection_params)
+    restore_command = build_restore_command(
+        extract_process, database, dump_filename, connection_params
+    )
 
     logger.debug(f"{log_prefix}: Restoring MongoDB database {database['name']}{dry_run_label}")
     if dry_run:
@@ -142,10 +146,16 @@ def build_restore_command(extract_process, database, dump_filename, connection_p
     '''
     Return the mongorestore command from a single database configuration.
     '''
-    hostname = connection_params['hostname'] or database.get('restore_hostname', database.get('hostname'))
+    hostname = connection_params['hostname'] or database.get(
+        'restore_hostname', database.get('hostname')
+    )
     port = str(connection_params['port'] or database.get('restore_port', database.get('port', '')))
-    username = connection_params['username'] or database.get('restore_username', database.get('username'))
-    password = connection_params['password'] or database.get('restore_password', database.get('password'))
+    username = connection_params['username'] or database.get(
+        'restore_username', database.get('username')
+    )
+    password = connection_params['password'] or database.get(
+        'restore_password', database.get('password')
+    )
 
     command = ['mongorestore']
     if extract_process:
--- a/borgmatic/hooks/mongodb.py
+++ b/borgmatic/hooks/mongodb.py
@@ -102,7 +102,7 @@ def make_database_dump_pattern(
     return dump.make_database_dump_filename(make_dump_path(location_config), name, hostname='*')
 
 
-def restore_database_dump(database_config, log_prefix, location_config, dry_run, extract_process):
+def restore_database_dump(database_config, log_prefix, location_config, dry_run, extract_process, connection_params):
     '''
     Restore the given MongoDB database from an extract stream. The database is supplied as a
     one-element sequence containing a dict describing the database, as per the configuration schema.
@@ -122,7 +122,7 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     dump_filename = dump.make_database_dump_filename(
         make_dump_path(location_config), database['name'], database.get('hostname')
     )
-    restore_command = build_restore_command(extract_process, database, dump_filename)
+    restore_command = build_restore_command(extract_process, database, dump_filename, connection_params)
 
     logger.debug(f"{log_prefix}: Restoring MongoDB database {database['name']}{dry_run_label}")
     if dry_run:
@@ -138,10 +138,15 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     )
 
 
-def build_restore_command(extract_process, database, dump_filename):
+def build_restore_command(extract_process, database, dump_filename, connection_params):
     '''
     Return the mongorestore command from a single database configuration.
     '''
+    hostname = connection_params['hostname'] or database.get('restore_hostname', database.get('hostname'))
+    port = str(connection_params['port'] or database.get('restore_port', database.get('port')))
+    username = connection_params['username'] or database.get('restore_username', database.get('username'))
+    password = connection_params['password'] or database.get('restore_password', database.get('password'))
+
     command = ['mongorestore']
     if extract_process:
         command.append('--archive')
@@ -149,14 +154,14 @@ def build_restore_command(extract_process, database, dump_filename):
         command.extend(('--dir', dump_filename))
     if database['name'] != 'all':
         command.extend(('--drop', '--db', database['name']))
-    if 'hostname' in database:
-        command.extend(('--host', database['hostname']))
-    if 'port' in database:
-        command.extend(('--port', str(database['port'])))
-    if 'username' in database:
-        command.extend(('--username', database['username']))
-    if 'password' in database:
-        command.extend(('--password', database['password']))
+    if hostname:
+        command.extend(('--host', hostname))
+    if port:
+        command.extend(('--port', str(port)))
+    if username:
+        command.extend(('--username', username))
+    if password:
+        command.extend(('--password', password))
     if 'authentication_database' in database:
         command.extend(('--authenticationDatabase', database['authentication_database']))
     if 'restore_options' in database:
--- a/borgmatic/hooks/postgresql.py
+++ b/borgmatic/hooks/postgresql.py
@@ -222,7 +222,7 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     database = database_config[0]
 
     hostname = connection_params['hostname'] or database.get('restore_hostname', database.get('hostname'))
-    port = str(connection_params['port'] or database.get('restore_port', database.get('port')))
+    port = str(connection_params['port'] or database.get('restore_port', database.get('port', '')))
     username = connection_params['username'] or database.get('restore_username', database.get('username'))
 
     all_databases = bool(database['name'] == 'all')
--- a/borgmatic/hooks/postgresql.py
+++ b/borgmatic/hooks/postgresql.py
@@ -23,13 +23,15 @@ def make_dump_path(location_config):  # pragma: no cover
     )
 
 
-def make_extra_environment(database):
+def make_extra_environment(database, restore=False):
     '''
     Make the extra_environment dict from the given database configuration.
     '''
     extra = dict()
     if 'password' in database:
         extra['PGPASSWORD'] = database['password']
+    if restore and 'restore_password' in database:
+        extra['PGPASSWORD'] = database['restore_password']
     extra['PGSSLMODE'] = database.get('ssl_mode', 'disable')
     if 'ssl_cert' in database:
         extra['PGSSLCERT'] = database['ssl_cert']
@@ -135,6 +137,7 @@ def dump_databases(databases, log_prefix, location_config, dry_run):
                 + (('--host', database['hostname']) if 'hostname' in database else ())
                 + (('--port', str(database['port'])) if 'port' in database else ())
                 + (('--username', database['username']) if 'username' in database else ())
+                + (('--no-owner',) if database['no_owner'] else ())
                 + (('--format', dump_format) if dump_format else ())
                 + (('--file', dump_filename) if dump_format == 'directory' else ())
                 + (tuple(database['options'].split(' ')) if 'options' in database else ())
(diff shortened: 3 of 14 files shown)


witten (commented):
Thanks for sharing your progress. Overall looks like a good approach. I left a few comments on the particulars. Looking forward to having this feature available for users!

borgmatic/hooks/postgresql.py
```diff
@@ -245,6 +245,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
 
     extra_environment = make_extra_environment(database)
 
+    if 'restore_password' in database:
+        extra_environment['PGPASSWORD'] = database['restore_password']
```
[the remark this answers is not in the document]
witten: Oh, good call! I totally didn't think of that. And yeah, adding an argument or a toggle for the restore use case would probably be a decent compromise.

borgmatic/hooks/postgresql.py
```diff
@@ -217,10 +217,10 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     analyze_command = (
         tuple(psql_command)
         + ('--no-password', '--no-psqlrc', '--quiet')
-        + (('--host', database['hostname']) if 'hostname' in database else ())
-        + (('--port', str(database['port'])) if 'port' in database else ())
-        + (('--username', database['username']) if 'username' in database else ())
-        + (('--dbname', database['name']) if not all_databases else ())
+        + (('--host', database.get('restore_hostname', database.get('hostname', ()))))
+        + (('--port', str(database.get('restore_port', database.get('port', ()))))
+        + (('--username', database.get('restore_username', database.get('username', ()))))
```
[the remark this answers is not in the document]
witten: Nope, I mean omitting both the username value _and_ the `--username` flag. That's what the previous code did when a `username` configuration option wasn't set.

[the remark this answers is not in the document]
diivi: Oh, got it!

borgmatic/actions/restore.py
```diff
@@ -68,15 +68,21 @@ def restore_single_database(
...
+    port, username and password as connection params,  and a configured database
     configuration dict, restore that database from the archive.
     '''
     logger.info(
         f'{repository.get("label", repository["path"])}: Restoring database {database["name"]}'
     )
 
+    logger.info(
+        f'hostname port username password for database {database["name"]}'
+    )
```
witten, line 84: I assume this is just a temporary debugging log?

witten (commented):
Overall approach looks good to me! I just had some feedback on some of the details.

borgmatic/actions/restore.py
```diff
@@ -308,6 +315,13 @@ def run_restore(
                 hooks, archive_database_names, hook_name, database_name
             )
 
+            connection_params = {
+                'hostname': restore_arguments.hostname,
+                'port': restore_arguments.port,
+                'username': restore_arguments.username,
+                'password': restore_arguments.password,
```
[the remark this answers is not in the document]
diivi: Well, no reason except restore arguments wasn't being used anywhere, so I thought this was a clear indication of what was happening. Can do either though.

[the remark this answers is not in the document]
witten: Up to you. I can see merits to both approaches.

borgmatic/hooks/postgresql.py
```diff
@@ -217,9 +230,9 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
     analyze_command = (
         tuple(psql_command)
         + ('--no-password', '--no-psqlrc', '--quiet')
-        + (('--host', database['hostname']) if 'hostname' in database else ())
-        + (('--port', str(database['port'])) if 'port' in database else ())
-        + (('--username', database['username']) if 'username' in database else ())
+        + (('--host', hostname) if 'hostname' in database or 'restore_hostname' in database or 'hostname' in connection_params else ())
+        + (('--port', port) if 'port' in database or 'restore_port' in database or 'port' in connection_params else ())
+        + (('--username', username) if 'username' in database or 'restore_username' in database or 'username' in connection_params else ())
```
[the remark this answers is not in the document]
diivi: Didn't get you 😅 Do you mean this:

```python
    hostname = connection_params['hostname'] or database.get('restore_hostname', database.get('hostname'))
```

Here, if everything is not provided, hostname should be none, and this piece of code (line 233) should exclude --host from the command,  right?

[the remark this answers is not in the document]
witten: Yup, that's the "logic above" I was referring to!

> Here, if everything is not provided, hostname should be none, and this piece of code (line 233) should exclude --host from the command, right?

Right. But wouldn't the following for line 233 accomplish that?

```python
(('--host', hostname) if hostname else ())
```

I might be missing something obvious here...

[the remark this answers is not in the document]
diivi: Oh lol, yeah you're right. I'll check and edit.

diivi:
@witten check element please (don't know how to ping you 😅)

witten:
That worked! (For pinging me.)

borgmatic/hooks/postgresql.py
```diff
@@ -22,14 +22,18 @@ def make_dump_path(location_config):  # pragma: no cover
...
     '''
     Make the extra_environment dict from the given database configuration.
+    If restore connection params are given, this is for a restore operation.
     '''
     extra = dict()
-    if 'password' in database:
-        extra['PGPASSWORD'] = database['password']
+
+    try:
+        extra['PGPASSWORD'] = restore_connection_params.get('password') or database['restore_password'] or database['password']
```
witten, line 33: I _think_ `database['restore_password']` needs to be `database.get('restore_password')` here, or you'll never get to `database['password']`.

Also, I'm pretty sure using the `'restore_password'` value from the `database` configuration needs to gate on whether `restore_connection_params` is non-`None`—since that's now your "flag" for whether this is a restore, right?

Soapbox time: Sometimes when I'm working on gnarly logic like this, I actually find it easier to write the tests _as_ I'm developing the function. That way, I can convince myself that the code I'm writing satisfies each use case. Doesn't always work (sometimes I still screw it up even after writing tests!), but sometimes it actually helps.

borgmatic/hooks/postgresql.py
```diff
@@ -202,13 +207,21 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
 
     If the extract process is None, then restore the dump from the filesystem rather than from an
     extract stream.
+
+    Use the given connection parameters to connect to the database. The connection parameters are
+    hostname, port, username, and password.
```
witten, line 212: It occurs to me you could use a `collections.namedtuple` for this data structure, just to formalize its contents a bit. Do not feel strongly at all.

witten (commented):
Overall lookin' good! I'm glad to see it's relatively straightforward to add similar functionality to the SQLite, MySQL, and MongoDB hooks.

borgmatic/hooks/mongodb.py
```diff
@@ -138,25 +138,30 @@ def restore_database_dump(database_config, log_prefix, location_config, dry_run,
...
-    if 'password' in database:
-        command.extend(('--password', database['password']))
+    if hostname:
+        command.extend(('--host', hostname))
+    if port:
+        command.extend(('--port', str(port)))
+    if username:
+        command.extend(('--username', username))
+    if password:
+        command.extend(('--password', password))
```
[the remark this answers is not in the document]
diivi: Yeah, I could do this in a future PR too (or get my friends to work on it, some of them are pretty enthusiastic to work on Borgmatic), since this pattern is followed throughout the file.

witten (commented):
You asked in chat about what other tests you might need. Here are some ideas:

 * The Postgres tests for `restore_database_dump()` are testing the password changes, but those changes are in a separate function—`make_extra_environment()`. So I'd instead recommend adding test coverage for your changes to `make_extra_environment()`, and then IMO you shouldn't need to include password-specific tests for the Postgres `restore_database_dump()`.
 * I would however recommend testing the username/hostname/port changes in the Postgres `restore_database_dump()`.
 * I'd also recommend unit testing all the permutations of username/port/etc. for `build_restore_command()` (or, where applicable, `restore_database_dump()`) in the other database hooks, since they've each got separate code to build their respective commands.
 
Let me know if you have any questions on any of this.

witten (commented):
And I know you're working on end-to-end tests as well. Those would be welcome, but they don't need to be exhaustive in terms of covering all the permutations—that's what unit tests are for IMO.

diivi:
@witten Matrix ping

tests/unit/hooks/test_postgresql.py
```diff
@@ -594,7 +624,233 @@ def test_restore_database_dump_runs_pg_restore_with_username_and_password():
...
+    database = {'name': 'foo', 'restore_password': 'trustsome1', 'password': 'anotherpassword'}
+
+    extra = module.make_extra_environment(
+        database, restore_connection_params={'password': 'clipassword'}
+    )
+
+    assert extra['PGPASSWORD'] == 'clipassword'
+
+
+def test_restore_database_dump_with_connection_params_uses_connection_params_for_restore():
```
diivi, line 651: Let me know what you think of these and I'll duplicate them for the other databases too!

tests/end-to-end/docker-compose.yaml
```diff
@@ -5,16 +5,38 @@ services:
...
       POSTGRES_PASSWORD: test
       POSTGRES_DB: test
+  postgresql2:
+    image: docker.io/postgres:13.1-alpine
+    environment:
+      POSTGRES_PASSWORD: test2
+      POSTGRES_DB: test
+      POSTGRES_USER: postgres2
+    ports:
+      - "5433:5432"
```
witten, line 15: I don't think this is being used. Especially since you're not connecting to this database on port 5433. You can probably remove this.

tests/end-to-end/docker-compose.yaml
```diff
@@ -5,16 +5,38 @@ services:
...
     environment:
       MYSQL_ROOT_PASSWORD: test
       MYSQL_DATABASE: test
+  mysql2:
+    image: docker.io/mariadb:10.5
+    environment:
+      MYSQL_ROOT_PASSWORD: test2
+      MYSQL_DATABASE: test
+    ports:
+      - "3307:3306"
```
witten, line 27: Same here.

tests/end-to-end/docker-compose.yaml
```diff
@@ -5,16 +5,38 @@ services:
...
     environment:
       MONGO_INITDB_ROOT_USERNAME: root
       MONGO_INITDB_ROOT_PASSWORD: test
+  mongodb2:
+    image: docker.io/mongo:5.0.5
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: root2
+      MONGO_INITDB_ROOT_PASSWORD: test2
+    ports:
+      - "27018:27017"
```
witten, line 39: And here.

tests/end-to-end/test_database.py
```diff
@@ -82,6 +82,108 @@ def write_configuration(
...
+
+hooks:
+    postgresql_databases:
+        - name: test
+          hostname: postgresql
+          username: postgres
+          password: test
+          format: {postgresql_dump_format}
+          restore_hostname: postgresql2
+          restore_port: 5432
```
witten, line 117: Note that this is the default port, so this won't actually override anything or test the restore port.

tests/end-to-end/test_database.py
```diff
@@ -82,6 +82,108 @@ def write_configuration(
...
+          restore_username: postgres2
+          restore_password: test2
+    mysql_databases:
+        - name: test
+          hostname: mysql
+          username: root
+          password: test
+          restore_hostname: mysql2
+          restore_port: 3306
```
witten, line 126: Same here.

tests/end-to-end/test_database.py
```diff
@@ -82,6 +82,108 @@ def write_configuration(
...
+          restore_password: test2
+    mongodb_databases:
+        - name: test
+          hostname: mongodb
+          username: root
+          password: test
+          authentication_database: admin
+          format: {mongodb_dump_format}
+          restore_hostname: mongodb2
+          restore_port: 27017
```
witten, line 137: And here!

tests/end-to-end/test_database.py
```diff
@@ -82,6 +82,108 @@ def write_configuration(
...
+        - name: sqlite_test
+          path: /tmp/sqlite_test.db
+          restore_path: /tmp/sqlite_test2.db
+'''
+
+    with open(config_path, 'w') as config_file:
+        config_file.write(config)
+
+
+def write_custom_restore_configuration_for_cli_arguments(
```
witten, line 150: This function totally makes sense to me, although I'm questioning its name a little bit.. Just because there's no code _in here_ specific to CLI arguments. So I wonder if it should have a less specific name and then just happen to be called from the CLI-testing test case below?

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,109 @@ def test_database_dump_and_restore():
...
 
 
+def test_database_dump_and_restore_with_restore_cli_arguments():
+    # Create a Borg repository.
+    temporary_directory = tempfile.mkdtemp()
+    repository_path = os.path.join(temporary_directory, 'test.borg')
+    borgmatic_source_directory = os.path.join(temporary_directory, '.borgmatic')
+
+    # Write out a special file to ensure that it gets properly excluded and Borg doesn't hang on it.
+    os.mkfifo(os.path.join(temporary_directory, 'special_file'))
```
witten, line 237: Not necessary to include here, IMO. This is testing a specific edge case that's important to test, but it doesn't need testing in _every_ e2e test case in this file as long as it's in one of them.

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,109 @@ def test_database_dump_and_restore():
...
+                '5432',
+                '--username',
+                'postgres2',
+                '--password',
+                'test2',
+            ]
+        )
+    finally:
+        os.chdir(original_working_directory)
+        shutil.rmtree(temporary_directory)
```
witten, line 287: This test looks good to me!

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,109 @@ def test_database_dump_and_restore():
...
+                '2',
+                '--config',
+                config_path,
+                'restore',
+                '--archive',
+                archive_name,
+                '--hostname',
+                'postgresql2',
+                '--port',
+                '5432',
```
witten, line 278: This also won't override anything, just because it's the default port.

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,109 @@ def test_database_dump_and_restore():
...
+                '--password',
+                'test2',
+            ]
+        )
+    finally:
+        os.chdir(original_working_directory)
+        shutil.rmtree(temporary_directory)
+
+
+def test_database_dump_and_restore_to_different_hostname_port_username_password():
```
witten, line 290: Might want to call this function something more generic like `test_database_dump_and_restore_with_restore_configuration_options()` for parity with the previous test case name.

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,109 @@ def test_database_dump_and_restore():
...
+
+        # Get the created archive name.
+        output = subprocess.check_output(
+            ['borgmatic', '--config', config_path, 'list', '--json']
+        ).decode(sys.stdout.encoding)
+        parsed_output = json.loads(output)
+
+        assert len(parsed_output) == 1
+        assert len(parsed_output[0]['archives']) == 1
+        archive_name = parsed_output[0]['archives'][0]['archive']
```
witten, line 322: I've heard it said before that test functions should be [DAMP not DRY](https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests). But you might consider factoring some of this out to a utility function anyway, just maybe something to get the created archive name since you're doing it in every test. Don't feel strongly.

tests/unit/hooks/test_postgresql.py
```diff
@@ -594,7 +624,233 @@ def test_restore_database_dump_runs_pg_restore_with_username_and_password():
...
+
+
+def test_make_extra_environment_with_cli_password_sets_correct_password():
+    database = {'name': 'foo', 'restore_password': 'trustsome1', 'password': 'anotherpassword'}
+
+    extra = module.make_extra_environment(
+        database, restore_connection_params={'password': 'clipassword'}
+    )
+
+    assert extra['PGPASSWORD'] == 'clipassword'
```
witten, line 648: What about the case where there's no CLI password and it's just a `restore_connection_params` password? Worth testing too?

tests/unit/hooks/test_postgresql.py
```diff
@@ -594,7 +624,233 @@ def test_restore_database_dump_runs_pg_restore_with_username_and_password():
...
+        connection_params={
+            'hostname': None,
+            'port': None,
+            'username': None,
+            'password': None,
+        },
+    )
+
+
+def test_restore_database_dump_with_no_passwords_runs_pg_restore_without_password():
```
witten, line 803: IMO this is an integration test because `make_extra_environment()` isn't mocked. But if you _do_ mock it, the test isn't really testing anything because all of the with/without password logic exists in `make_extra_environment()`. So my recommendation would be either to cut this test or, if you want to keep it, move it into `integration/`.

I know this unit/integration distinction might seem silly, but it goes back to that test pyramid.. It keeps us honest about what kinds of tests we're actually writing.

witten (commented):
Overall, tests seem good! I left a few notes here and there, but I think the general approach could be used for the other database hooks. The main difference of course is SQLite, since you won't be spinning up services for that one.

tests/unit/hooks/test_postgresql.py
diivi: restore_connection_params password is the password we get through the CLI though. Did you mean no config password?

tests/unit/hooks/test_postgresql.py
diivi: Ah okay, I have no issues cutting this out if it doesn't add value :D

tests/end-to-end/docker-compose.yaml
diivi: I think I should keep this and try to run postgres on the container's port 5433, but I wasn't able to do that. I found this - https://stackoverflow.com/questions/37775702/changing-a-postgres-containers-server-port-in-docker-compose but  I don't think it worked. I want to do 5433 and map it to the host's 5433 so that I can test the --restore option without the default, as you mention below.

diivi: Oh, the answer I linked seems to do what I want for postgres, I'll look for solutions for the other databases too.

diivi: Okay done for all, let me know if this is better.

tests/unit/hooks/test_postgresql.py
witten: Yes, my bad. That's what I meant!

witten: Actually, looking at the function under test, I think I meant: What about the case where there's no CLI password and it's just a config password?

tests/end-to-end/docker-compose.yaml
witten: So I think your `command: -p 5433` is a good fix here, because it means the second Postgres is running on a different port than the first one, so you can test the port overriding feature of your new code. But I still don't think the `ports: - "5433:5432"` is doing anything for your tests, because they're going through the container network rather than the host network.. You don't need to expose the port at all. (Unless I'm sorely mistaken..)

tests/unit/hooks/test_postgresql.py
diivi: Oh yeah, I remember now, I left it out because I thought `test_restore_database_dump_runs_pg_restore_with_username_and_password` tests the same thing doesn't it?

tests/end-to-end/docker-compose.yaml
diivi: Yup, it worked without the ports: thing, thanks!

tests/unit/hooks/test_postgresql.py
witten: Sorry, I'm being totally unclear. I mean the case where there's no CLI password and there's only a _restore_ config password.. (Or a restore config password + a standard `password`.)

tests/unit/hooks/test_sqlite.py
```diff
@@ -94,12 +95,79 @@ def test_restore_database_dump_restores_database():
...
+        'test.yaml',
+        {},
+        dry_run=False,
+        extract_process=extract_process,
+        connection_params={'restore_path': None},
+    )
+
+
+def test_restore_database_dump_with_connection_params_uses_connection_params_for_restore():
+    database_config = [{'path': '/path/to/database', 'name': 'database'}]
```
witten, line 121: Maybe include `restore_path` here to make sure it doesn't get used?

witten (commented):
Tests are looking great! I left a few minor comments/questions, but once those are resolved one way or the other, I think we're good to merge.

tests/unit/hooks/test_mongodb.py
```diff
@@ -253,7 +283,125 @@ def test_restore_database_dump_runs_mongorestore_with_username_and_password():
...
+def test_restore_database_dump_with_connection_params_uses_connection_params_for_restore():
+    database_config = [
+        {
+            'name': 'foo',
+            'username': 'mongo',
+            'password': 'trustsome1',
+            'authentication_database': 'admin',
+            'schemas': None,
+        }
+    ]
```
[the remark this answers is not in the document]
diivi: True, will do.

tests/end-to-end/test_database.py
```diff
@@ -125,6 +227,103 @@ def test_database_dump_and_restore():
...
+                '--hostname',
+                'postgresql2',
+                '--port',
+                '5433',
+                '--username',
+                'postgres2',
+                '--password',
+                'test2',
+            ]
+        )
```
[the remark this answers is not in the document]
witten: With the original database tests, there's only a single database instance per hook, so if the restore executes without erroring, you're pretty well assured that it worked. However with these new tests, there are two database instances per hook. So a silent failure mode is if a restore happens successfully—just to the wrong database instance. Like if all the `restore_*` options or new CLI flags get ignored.

One way to check for that would just be to run a query (`psql` or similar) against the expected database instance and make sure the named database in question actually exists there. If it does, then you know it must've been because you restored it. (But that will only work if the `POSTGRES_DATABASE` in the `docker-compose.yaml` is different for each database instance.)

An argument against doing this is that you've already got coverage for the "wrong database instance" in your unit tests, so maybe you don't need it / want it in your end-to-end tests too. Ultimately, I'll leave the decision to you. Just raising the question here in case you want to do anything about it.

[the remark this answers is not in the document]
diivi: Hm, thanks for letting me know. 

Also, won't docker create the POSTGRES_DB in docker_compose.yaml as soon as I start the tests (even before backup)? If yes, even the database check query won't work and I'll have to check for something else inside the database.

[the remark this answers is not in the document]
witten: Yeah, that's why I was suggesting two different `POSTGRES_DB` names, one for each database instance. Like if the first instance's database is `test1` and the second instance's is `test2`. Then if you attempt to restore `test1` to the second instance, and you see `test1` actually created there, you know it must've worked.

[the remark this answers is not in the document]
diivi: Doesn't the restore command already need the database to exist?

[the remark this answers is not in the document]
diivi: https://www.postgresql.org/docs/current/app-pgrestore.html I think that's what the --create option is for.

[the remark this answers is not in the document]
witten: You are correct! Without that flag, it won't create it. You could _add_ the flag though, which might not be a bad idea to have anyway.. Assuming it still works when the database already exists.

witten:
Talking about scope creep, this question just came in on IRC:

> is there a way to restore a particular MariaDB Database to a new name?

So not as part of this PR, but maybe as follow-up work, what do you think about adding a `restore_name` option and corresponding `borgmatic restore --restore-database` flag? Seems useful. I've given zero thought as to how each that'd be to implement though.

witten:
Thanks for all of your hard work here!!!

diivi:
> Talking about scope creep, this question just came in on IRC:
> 
> > is there a way to restore a particular MariaDB Database to a new name?
> 
> So not as part of this PR, but maybe as follow-up work, what do you think about adding a `restore_name` option and corresponding `borgmatic restore --restore-database` flag? Seems useful. I've given zero thought as to how easy that'd be to implement though.

I'll have a look around the end of GSoC, I remember trying it some time ago but I got stuck somewhere.

tests/end-to-end/test_database.py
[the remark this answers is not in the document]
witten: Just following up on this. It turns out that restores are [not working correctly in this end-to-end test](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/738#issuecomment-6912) in main—likely broken by me _after_ you submitted this PR. But my point in bringing this up now is that more assertions on whether the tests were actually restoring the expected data probably would've caught that!

PR #74, opened 2023-07-15
add docs for database restore params and config bootstrap

diivi:
image
image

### DIFF
--- a/docs/how-to/backup-your-databases.md
+++ b/docs/how-to/backup-your-databases.md
@@ -339,10 +339,10 @@ borgmatic restore --archive latest --database users --schema tentant1
 ```
 
 ### Restore to an alternate host
-<span class="minilink minilink-addedin">New in version 1.7.15</span> You
-can restore a database dump to a host other than the one from which it was
-originally dumped. You can also change other connection parameters like the
-username, password, and port. This can be done from the command line:
+<span class="minilink minilink-addedin">New in version 1.7.15</span>
+A database dump can be restored to a host other than the one from which it was
+originally dumped. The connection parameters like the username, password, and
+port can also be changed. This can be done from the command line:
 
 ```bash
 borgmatic restore --archive latest --database users --hostname database2.example.org --port 5433 --username postgres --password trustsome1
@@ -351,14 +351,13 @@ borgmatic restore --archive latest --database users --hostname database2.example
 Or from the configuration file:
 
 ```yaml
-hooks:
-    postgresql_databases:
-        - name: users
-          hostname: database1.example.org
-          restore_hostname: database1.example.org
-          restore_port: 5433
-          restore_username: postgres
-          restore_password: trustsome1
+postgresql_databases:
+    - name: users
+        hostname: database1.example.org
+        restore_hostname: database1.example.org
+        restore_port: 5433
+        restore_username: postgres
+        restore_password: trustsome1
 ```
 
 
--- a/docs/how-to/backup-your-databases.md
+++ b/docs/how-to/backup-your-databases.md
@@ -338,6 +338,29 @@ within the database dump:
 borgmatic restore --archive latest --database users --schema tentant1
 ```
 
+### Restore to an alternate host
+<span class="minilink minilink-addedin">New in version 1.7.15</span> You
+can restore a database dump to a host other than the one from which it was
+originally dumped. You can also change other connection parameters like the
+username, password, and port. This can be done from the command line:
+
+```bash
+borgmatic restore --archive latest --database users --hostname database2.example.org --port 5433 --username postgres --password trustsome1
+```
+
+Or from the configuration file:
+
+```yaml
+hooks:
+    postgresql_databases:
+        - name: users
+          hostname: database1.example.org
+          restore_hostname: database1.example.org
+          restore_port: 5433
+          restore_username: postgres
+          restore_password: trustsome1
+```
+
 
 ### Limitations
 
--- a/docs/how-to/extract-a-backup.md
+++ b/docs/how-to/extract-a-backup.md
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
 `--destination` flag.
 
 
+## Extract the configuration file used to create an archive
+
+<span class="minilink minilink-addedin">New in version 1.7.15</span> Borgmatic
+also stores the configuration file used to create an archive, inside the
+archive itself. This is useful in cases where you've lost your configuration
+file, or you want to see what configuration was used to create a particular
+archive.
+
+To extract the configuration file from an archive, use the `config bootstrap` action:
+
+```bash 
+borgmatic config bootstrap --repository repo.borg --destination /tmp
+```
+
+This extracts the configuration file from the latest archive in the repository `repo.borg` to `/tmp/config.yaml`.
+
 ## Repository selection
 
 If you have a single repository in your borgmatic configuration file(s), no
--- a/docs/how-to/extract-a-backup.md
+++ b/docs/how-to/extract-a-backup.md
@@ -150,7 +150,7 @@ borgmatic umount --mount-point /mnt
 <span class="minilink minilink-addedin">New in version 1.7.15</span> borgmatic
 automatically stores all the configuration files used to create an archive inside the
 archive itself. This is useful in cases where you've lost a configuration
-file, or you want to see what configurations were used to create a particular
+file or you want to see what configurations were used to create a particular
 archive.
 
 To extract the configuration files from an archive, use the `config bootstrap` action. For example:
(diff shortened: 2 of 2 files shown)


docs/how-to/backup-your-databases.md
```diff
@@ -338,6 +338,29 @@ within the database dump:
 borgmatic restore --archive latest --database users --schema tentant1
 ```
 
+### Restore to an alternate host
+<span class="minilink minilink-addedin">New in version 1.7.15</span> You
+can restore a database dump to a host other than the one from which it was
+originally dumped. You can also change other connection parameters like the
+username, password, and port. This can be done from the command line:
```
witten, line 345: Nit: The info here is good, but just from an editing prose perspective, there's lots of "you can" and "can be done" going on in this paragraph. A little of that is totally fine of course; I'm just calling out the repeated use.

docs/how-to/backup-your-databases.md
```diff
@@ -338,6 +338,29 @@ within the database dump:
...
+
+```bash
+borgmatic restore --archive latest --database users --hostname database2.example.org --port 5433 --username postgres --password trustsome1
+```
+
+Or from the configuration file:
+
+```yaml
+hooks:
```
witten, line 354: `hooks:` is gone in main!

docs/how-to/backup-your-databases.md
```diff
@@ -338,6 +338,29 @@ within the database dump:
...
+    postgresql_databases:
+        - name: users
+          hostname: database1.example.org
+          restore_hostname: database1.example.org
+          restore_port: 5433
+          restore_username: postgres
+          restore_password: trustsome1
+```
```
witten, line 362: Overall this section looks great to me!

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
 `--destination` flag.
 
 
+## Extract the configuration file used to create an archive
+
+<span class="minilink minilink-addedin">New in version 1.7.15</span> Borgmatic
```
witten, line 50: Nit: borgmatic is lowercased (even at the start of sentences!).

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
 `--destination` flag.
 
 
+## Extract the configuration file used to create an archive
```
witten, line 48: I feel like this is at a odd spot in flow of the page, even though I think I see why you put it first-ish: Because `bootstrap` needs to come first on a fresh system. Anyway, the previous section is about `extract`ing individual files from an existing repo, and the section after this new one is about a variant of that.. specifying a single repository to use with `extract`. Neither of which have to do with bootstrapping!

So here's an alternate idea on where to put your new section such that the whole page flows a little better:

 * Start the page with a brief intro paragraph: "When the worst happens ..." etc etc.
 * Then insert your new section, perhaps prefaced with: "If you're extracting to a fresh system that doesn't have borgmatic configuration files ..." or whatever.
 * And then the next section can dive into the existing `extract` docs with another preface: "If you already have borgmatic configuration files (or you've just bootstrapped them as per above), you can extract your files from a repository. A good way to do that is to use the `rlist` action ..."

If you don't like that approach, then maybe just leave the document mostly alone and then move your new section to the end? That way it doesn't interrupt the flow of talking about the `extract` action across sections?

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
...
+
+<span class="minilink minilink-addedin">New in version 1.7.15</span> Borgmatic
+also stores the configuration file used to create an archive, inside the
```
witten, line 51: Nit: Remove comma.

witten, line 51: Maybe change "also" to "automatically"?

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
...
+archive itself. This is useful in cases where you've lost your configuration
+file, or you want to see what configuration was used to create a particular
+archive.
+
+To extract the configuration file from an archive, use the `config bootstrap` action:
```
witten, line 56: In this whole section, is it worth pluralizing this and talking about extracting "configuration file**s**" wherever they're mentioned? That seems like an important (and cool!) part of the feature.

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
...
+
+```bash 
+borgmatic config bootstrap --repository repo.borg --destination /tmp
+```
```
witten, line 60: I think it's also worth mentioning that:

1. You don't need any existing configuration files on disk in order to use the `bootstrap` feature! (That's why `--repository` is required.)
2. If you don't specify a destination, it will extract configuration files to their original path, silently overwriting any configuration files that may already exist. I think that's important to mention both because it seems like a common way to use the feature and also because it might be a little surprising if someone isn't expecting the overwrite.
3. IMO not all command-line flags of an action need to be documented, but it seems like `--archive` is important to mention..?

docs/how-to/extract-a-backup.md
```diff
@@ -45,6 +45,22 @@ in the right place before running the command—or see below about the
...
+
+This extracts the configuration file from the latest archive in the repository `repo.borg` to `/tmp/config.yaml`.
```
witten, line 62: Wouldn't it be more like `/tmp/etc/borgmatic/config.yaml`? (You could explicitly mention that it includes the directory components of its original path or just let the example speak for itself.) Also, this last sentence should probably start with "For instance," or similar, as the user may very well have different configuration file(s) with different paths.

witten (changes requested):
Overall this is looking good, especially the database changes. I added several suggestions to the bootstrap section though; I feel like it could probably use a few more details in places. Anyway, thank you for doing these docs!

docs/how-to/extract-a-backup.md
witten: Thanks for pointing that out.. fixed!

witten: And lemme know when you push your changes and you're ready for review again!

docs/how-to/extract-a-backup.md
diivi: Moved this section to the end since I didn't wanna move a lot of stuff around here.

docs/how-to/extract-a-backup.md
```diff
@@ -145,3 +144,29 @@ When you're all done exploring your files, unmount your mount point. No
...
+
+```bash 
+borgmatic config bootstrap --repository repo.borg --destination /tmp
+```
+
+This extracts the configuration file from the latest archive in the repository `repo.borg` to `/tmp/etc/borgmatic/config.yaml`, assuming that the only configuration file used to create this archive was located at `/etc/borgmatic/config.yaml` when the archive was created.
+
+Note that to run the `config bootstrap` action, you don't need to have a borgmatic configuration file. You only need to specify the repository to use via the `--repository` flag, borgmatic will figure out the rest.
+
+If a destination directory is not specified, the configuration files will be extracted to their original locations, silently **overwriting** any configuration files that may already exist. For example, if a configuration file was located at `/etc/borgmatic/config.yaml` when the archive was created, it will be extracted to `/etc/borgmatic/config.yaml` too. This is in line with the `extract` action, which extracts files to their original locations by default.
```
witten, line 166: This last sentence isn't actually true! `extract` extracts to the current directory by default. Maybe just cut that part?

docs/how-to/extract-a-backup.md
```diff
@@ -145,3 +144,29 @@ When you're all done exploring your files, unmount your mount point. No
...
+
+To extract the configuration files from an archive, use the `config bootstrap` action. For example:
+
+```bash 
+borgmatic config bootstrap --repository repo.borg --destination /tmp
+```
+
+This extracts the configuration file from the latest archive in the repository `repo.borg` to `/tmp/etc/borgmatic/config.yaml`, assuming that the only configuration file used to create this archive was located at `/etc/borgmatic/config.yaml` when the archive was created.
+
+Note that to run the `config bootstrap` action, you don't need to have a borgmatic configuration file. You only need to specify the repository to use via the `--repository` flag, borgmatic will figure out the rest.
```
witten, line 164: Grammar nit: Maybe change that last comma to a semi-colon? Alternatively, an em dash would work too.

docs/how-to/extract-a-backup.md
```diff
@@ -145,3 +144,29 @@ When you're all done exploring your files, unmount your mount point. No
 ```bash
 borgmatic umount --mount-point /mnt
 ```
+
+## Extract the configuration files used to create an archive
+
+<span class="minilink minilink-addedin">New in version 1.7.15</span> borgmatic
+automatically stores all the configuration files used to create an archive inside the
+archive itself. This is useful in cases where you've lost a configuration
+file, or you want to see what configurations were used to create a particular
```
witten, line 153: Grammar nit: Remove comma.

witten (changes requested):
Looks great. A couple minor tweaks and then I think we're done!

docs/how-to/extract-a-backup.md
diivi, line 164:
```suggestion
Note that to run the `config bootstrap` action, you don't need to have a borgmatic configuration file. You only need to specify the repository to use via the `--repository` flag; borgmatic will figure out the rest.
```

docs/how-to/extract-a-backup.md
diivi, line 166:
```suggestion
If a destination directory is not specified, the configuration files will be extracted to their original locations, silently **overwriting** any configuration files that may already exist. For example, if a configuration file was located at `/etc/borgmatic/config.yaml` when the archive was created, it will be extracted to `/etc/borgmatic/config.yaml` too.
```

docs/how-to/extract-a-backup.md
diivi, line 153:
```suggestion
file or you want to see what configurations were used to create a particular
```

witten:
Thank you!

PR #75, opened 2023-07-31
feat: optionally disable config bootstrap

diivi:
[#725](https://projects.torsion.org/borgmatic-collective/borgmatic/issues/725)

### DIFF
--- a/tests/unit/borg/test_create.py
+++ b/tests/unit/borg/test_create.py
@@ -651,13 +651,13 @@ def test_create_archive_with_sources_and_used_config_paths_with_store_config_fil
     flexmock(module.logging).ANSWER = module.borgmatic.logger.ANSWER
     flexmock(module).should_receive('collect_borgmatic_source_directories').and_return([])
     flexmock(module).should_receive('deduplicate_directories').and_return(
-        ('foo', 'bar', '/etc/borgmatic/config.yaml')
+        ('foo', 'bar')
     )
     flexmock(module).should_receive('map_directories_to_devices').and_return({})
     flexmock(module).should_receive('expand_directories').with_args([]).and_return(())
     flexmock(module).should_receive('expand_directories').with_args(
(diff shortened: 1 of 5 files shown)


diivi:
Nevermind, it still stored the config 🤔

witten (changes requested):
Looks great, but would you also mind adding some text coverage for this? It could be a very basic unit test that exercises the non-default code path. Thank you!

witten:
> Nevermind, it still stored the config

Huh, weird. This totally looks like it'd work. Maybe you're not using the latest archive when looking for the stored config?

witten:
Oh, maybe this changeset only prevents storage of the manifest but doesn't change whether the actual config files get stored?

diivi:
> Oh, maybe this changeset only prevents storage of the manifest but doesn't change whether the actual config files get stored?

Yup

> Looks great, but would you also mind adding some test coverage for this? It could be a very basic unit test that exercises the non-default code path. Thank you!

On it!

tests/unit/borg/test_create.py
```diff
@@ -646,6 +646,53 @@ def test_create_archive_with_sources_and_used_config_paths_calls_borg_with_sourc
         global_arguments=flexmock(log_json=False, used_config_paths=['/etc/borgmatic/config.yaml']),
     )
 
+def test_create_archive_with_sources_and_used_config_paths_with_store_config_files_false_calls_borg_with_sources_and_no_config_paths():
```
diivi, line 649: Disclaimer: this is Copilot generated. I just took a glance (lines 691 and 677) and it made sense + it passed.

witten:
The test looks fine to me (even though that function name makes even _me_ blush), but what about a test for the manifest code path?

diivi:
Done!

witten:
Awesome, thank you!

PR #77, opened 2023-08-05
Fix typos.

hop:
Some typos and inconsistencies I found while browsing the sample configuration.

### DIFF
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -7,7 +7,7 @@ properties:
         type: object
         description: |
             Constants to use in the configuration file. All occurrences of the
-            constant name within culy braces will be replaced with the value.
+            constant name within curly braces will be replaced with the value.
             For example, if you have a constant named "hostname" with the value
             "myhostname", then the string "{hostname}" will be replaced with
             "myhostname" in the configuration file.
@@ -19,8 +19,8 @@ properties:
         items:
             type: string
         description: |
-            List of source directories and files to backup. Globs and tildes are
-            expanded. Do not backslash spaces in path names.
+            List of source directories and files to back up. Globs and tildes
+            are expanded. Do not backslash spaces in path names.
         example:
             - /home
             - /etc
@@ -29,7 +29,7 @@ properties:
     repositories:
         type: array
         items:
-            type: object 
+            type: object
             required:
                 - path
             properties:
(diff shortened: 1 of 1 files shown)


borgmatic/config/schema.yaml
```diff
@@ -1355,8 +1355,8 @@ properties:
                     - finish
         description: |
             Configuration for a monitoring integration with Healthchecks. Create
-            an account at https://healthchecks.io (or self-host Healthchecks) if
-            you'd like to use this service. See borgmatic monitoring
+            an account at https://healthchecks.io/ (or self-host Healthchecks)
```
witten, line 1358: Is there a reason you've added trailing slashes for these URLs when the the canonical versions don't appear to have trailing slashes?

Some related reading: https://ahrefs.com/blog/trailing-slash/

borgmatic/config/schema.yaml
```diff
@@ -19,8 +19,8 @@ properties:
         items:
             type: string
         description: |
-            List of source directories and files to backup. Globs and tildes are
-            expanded. Do not backslash spaces in path names.
+            List of source directories and files to back up. Globs and tildes
```
witten, line 22: This change is correct, but I'll admit I don't want it to be!

witten (commented):
Wow, you win the daily conscientiousness award for sure. :trophy: Thanks for taking the time to fix these. I just have one question about the trailing slash changes. (See comments.)

borgmatic/config/schema.yaml
hop: TBH, this is just personal preference from decades of writing Apache configs and stuff like that. Feel free to ignore!

borgmatic/config/schema.yaml
hop: Yeah, I was torn about this one… I just included it because "backing up" and "backed up" is used elsewhere.

witten: Totally makes sense.

borgmatic/config/schema.yaml
witten: Understood. I think I'd prefer to use the canonical URLs for these projects rather than inserting a trailing slash—even if it does look nicer. So would you mind taking these out again? And then I'd be happy to merge the PR so as to get the other fixes!

hop: Done.

I also remove one such slash that was there before.

witten:
Awesome, thank you!

PR #78, opened 2023-10-15
Update systemd .service example

Alphix:
First, ProtectSystem=strict will make the entire file system hierarchy (except /dev, /proc/ and /sys) read-only, so separate ReadOnlyPaths= is not necessary.

Second, ProtectHome=tmpfs will not just mount an empty tmpfs on /root, but also on /home and /run/user. As it's likely quite common to want to backup /home, this seems like a footgun.

Finally, it's quite likely that borgbackup will want access to root's SSH keys in order to connect to remote backup servers.

Note that all these options are commented out by default, so this is more of a documentation change than any real change in functionality.

### DIFF
--- a/sample/systemd/borgmatic.service
+++ b/sample/systemd/borgmatic.service
@@ -32,16 +32,16 @@ RestrictSUIDSGID=yes
 SystemCallArchitectures=native
 SystemCallFilter=@system-service
 SystemCallErrorNumber=EPERM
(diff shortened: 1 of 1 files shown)


witten:
Awesome, thanks so much for taking the time to do this and submit a PR!

Alphix:
> Awesome, thanks so much for taking the time to do this and submit a PR!

Pleasure doing business with ya guv, thanks for the quick turnaround ;)

PR #79, opened 2023-10-27
Added CODE_OF_CONDUCT.md

parthrc:
Added a CODE_OF_CONDUCT filebased on the Contributor Covenant for better contributor experience.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,127 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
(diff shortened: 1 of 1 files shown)


witten:
Thanks for taking the time to submit this! I like the general idea of formalizing our community's code of conduct. A couple of thoughts:

 * Is there a reason you selected 2.0 of the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct/) instead of 2.1?
 * I'm not sure I agree with this part as phrased: "Use of inappropriate language or other behavior deemed
unprofessional or unwelcome in the community." My rationale is that "inappropriate" can be interpreted as "foul," and there is definitely occasional swearing on borgmatic tickets that I don't think should violate any code of conduct. Maybe the word "derogatory" instead of "inappropriate" would be a fix for this concern?
 * I'm not wild about the slow escalation from correction to warning to temporary ban to permanent ban. When a spammer abuses the project site, I often resort to an immediate permanent ban. I wouldn't want to give up that ability, because this tiny open source project doesn't have the staff to baby users through multiple different penalty boxes.
 * Speaking of, I'd want spamming / abusing project resources for commercial interests to be mentioned as something not to do!

parthrc:
Hey @witten thanks for the feedback!

- the 2.1 version has only 2 words added which although less should be considered, which I was not aware.
- As for your three suggestions, absolutely anything you want should be added to this file because ultimately you are the maintainer. I added this file as a baseline for what can be done.
- I can make the changes you mentioned and update it, then we can discuss further.

witten:
That would be great.. Thank you for your flexibility here!

PR #81, opened 2024-03-02
Custom dump command for mysql

shivansh02:
Related Issue:
[https://projects.torsion.org/borgmatic-collective/borgmatic/issues/311](url)

Motivation and Context
Lets you run custom dump and restore commands for mysql to specify path for dump/restore binaries.

How Has This Been Tested?
Tested by me locally. I am able to run custom commands.

Types of change: New Feature

I have added tests to cover my changes.

### DIFF
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -1092,15 +1092,15 @@ properties:
                     description: |
                         Command to use instead of "mysqldump". This can be used
                         to run a specific mysql_dump version (e.g., one inside 
-                        a running docker container). Defaults to "mysqldump".
+                        a running container). Defaults to "mysqldump".
                     example: docker exec mysql_container mysqldump
                 mysql_command:
                     type: string
                     description: |
                         Command to run instead of "mysql". This
                         can be used to run a specific mysql
-                        version (e.g., one inside a running docker
-                        container). Defaults to "mysql".
+                        version (e.g., one inside a running container). 
+                        Defaults to "mysql".
                     example: docker exec mysql_container mysql
                 format:
                     type: string
--- a/borgmatic/config/schema.yaml
+++ b/borgmatic/config/schema.yaml
@@ -971,6 +971,22 @@ properties:
                         a password will only work if MariaDB is configured to
                         trust the configured username without a password.
                     example: trustsome1
+                mariadb_dump_command:
+                    type: string
+                    description: |
+                        Command to use instead of "mariadb-dump". This can 
+                        be used to run a specific mariadb_dump version 
+                        (e.g., one inside a running container). 
+                        Defaults to "mariadb-dump".
+                    example: docker exec mariadb_container mariadb-dump
+                mariadb_command:
+                    type: string
+                    description: |
+                        Command to run instead of "mariadb". This
+                        can be used to run a specific mariadb
+                        version (e.g., one inside a running container). 
+                        Defaults to "mariadb".
+                    example: docker exec mariadb_container mariadb           
                 restore_password:
                     type: string
                     description: |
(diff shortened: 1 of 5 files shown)


borgmatic/config/schema.yaml
```diff
@@ -1087,6 +1087,21 @@ properties:
...
+                        to run a specific mysql_dump version (e.g., one inside 
+                        a running docker container). Defaults to "mysqldump".
+                    example: docker exec mysql_container mysqldump
+                mysql_command:
+                    type: string
+                    description: |
+                        Command to run instead of "mysql". This
+                        can be used to run a specific mysql
+                        version (e.g., one inside a running docker
+                        container). Defaults to "mysql".
```
witten, line 1103: Would you mind wrapping this to 80 characters? Thanks.

borgmatic/config/schema.yaml
```diff
@@ -1087,6 +1087,21 @@ properties:
                         Password with which to connect to the restore database.
                         Defaults to the "password" option.
                     example: trustsome1
+                mysql_dump_command:
+                    type: string
+                    description: |
+                        Command to use instead of "mysqldump". This can be used
+                        to run a specific mysql_dump version (e.g., one inside 
+                        a running docker container). Defaults to "mysqldump".
```
witten, line 1095: IMO, remove "docker" in the description (but not the example) since this isn't specific to Docker containers. E.g. it applies to Podman as well.

borgmatic/hooks/mysql.py
```diff
@@ -79,8 +79,9 @@ def execute_dump_command(
         )
         return None
 
+    mysql_dump_command = database.get('mysql_dump_command') or 'mysqldump'
```
witten, line 82: To prevent shell injections attacks, this (and all custom commands in `mysql.py`) need to be escaped. See the `postgresql.py` hook (like on like 77) for an example.

borgmatic/hooks/mysql.py
```diff
@@ -206,9 +207,9 @@ def restore_data_source_dump(
     password = connection_params['password'] or data_source.get(
         'restore_password', data_source.get('password')
     )
-
+    mysql_restore_command = data_source.get('mysql_command') or 'mysql'
     restore_command = (
-        ('mysql', '--batch')
+        (mysql_restore_command, '--batch')
```
witten, line 212: If I'm reading this correctly, you're missing one other instance of `mysql`/`mysqldump` in `mysql.py`: Around line 39 in `database_names_to_dump()`.

witten (changes requested):
First of all, thank you for taking the time to submit a PR for this! And tests too.. awesome. The overall approach looks good, although I added a few comments requesting some hopefully smaller changes.

Also, while you're here, would you mind doing the `mariadb.py` hook and config as well? It's almost identical to the `mysql.py` hook, so it _should_ just be a search and replace or close to it.

borgmatic/config/schema.yaml
shivansh02: I'm not sure I follow. It appears to be under 80 characters per line already.

shivansh02:
Thanks for the review @witten ! Made all the requested changes, except the one about schema formatting. Also added custom commands for MariaDB. Please take a look.

borgmatic/config/schema.yaml
witten: It is less than 80, but it looks like these lines are wrapped to a _smaller_ character count, e.g. 70.

shivansh02: Oh my bad! I thought it was good as long as it was under 80.

witten: No worries! It's not a big deal either way.

PR #82, opened 2024-03-04
support for NO_COLOR env var

shivansh02:
Solves [https://projects.torsion.org/borgmatic-collective/borgmatic/issues/835](url)

Motivation
see [https://no-color.org/](url)

I have added/modified tests to cover my changes.

### DIFF
--- a/borgmatic/logger.py
+++ b/borgmatic/logger.py
@@ -42,7 +42,7 @@ def should_do_markup(no_color, configs):
         return False
 
     no_color_env = os.environ.get('NO_COLOR', None)
-    if no_color_env is not None and to_bool(no_color_env):
+    if no_color_env is not None:
         return False
 
     py_colors = os.environ.get('PY_COLORS', None)
--- a/borgmatic/logger.py
+++ b/borgmatic/logger.py
@@ -41,6 +41,10 @@ def should_do_markup(no_color, configs):
     if any(config.get('output', {}).get('color') is False for config in configs.values()):
         return False
 
+    no_color_env = os.environ.get('NO_COLOR', None)
+    if no_color_env is not None and to_bool(no_color_env):
+        return False
+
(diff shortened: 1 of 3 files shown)


borgmatic/logger.py
```diff
@@ -41,6 +41,10 @@ def should_do_markup(no_color, configs):
     if any(config.get('output', {}).get('color') is False for config in configs.values()):
         return False
 
+    no_color_env = os.environ.get('NO_COLOR', None)
+    if no_color_env is not None and to_bool(no_color_env):
```
witten, line 45: I don't think this is quite right if I'm reading the [specification](https://no-color.org/) correctly. It sounds like `NO_COLOR` is considered "set" (true) whenever it has *any* value—not just when it has a boolean value that evaluates to true. From the spec:

> Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (regardless of its value), prevents the addition of ANSI color.

witten (changes requested):
Overall approach looks good, but I left a comment about interpreting the actual value of `NO_COLOR`. Once that's addressed, I think this should be good to merge.

Thank you!

borgmatic/logger.py
shivansh02: Oh yeah you're right, I'll fix it.

borgmatic/logger.py
```diff
@@ -42,7 +42,7 @@ def should_do_markup(no_color, configs):
         return False
 
     no_color_env = os.environ.get('NO_COLOR', None)
-    if no_color_env is not None and to_bool(no_color_env):
+    if no_color_env is not None:
```
witten, line 45: Super minor, but this could easily be on one line now (cutting out the intermediate variable).

Do not feel strongly.

witten (approved):
LGTM! Thanks for making the requested change from the last review.

PR #84, opened 2024-08-08
WIP:MySQL Directory Format

shivansh02:
Adds support for directory format for MySQL databases.

To Do:
- [ ] Write new/modify unit tests
- [ ] Modify E2E tests

### DIFF
--- a/borgmatic/hooks/mysql.py
+++ b/borgmatic/hooks/mysql.py
@@ -87,8 +87,7 @@ def execute_dump_command(
         shlex.quote(part) for part in shlex.split(database.get('mysql_dump_command') or 'mysqldump')
     )
 
-    dump_format = database.get('format')
-    is_directory_format = dump_format == 'directory'
+    is_directory_format = database.get('format') == 'directory'
 
     dump_command = (
         mysql_dump_command
@@ -116,13 +115,12 @@ def execute_dump_command(
             extra_environment=extra_environment,
         )
         return None
-    else:
-        dump.create_named_pipe_for_dump(dump_filename)
-        return execute_command(
-            dump_command,
-            extra_environment=extra_environment,
-            run_to_completion=False,
-        )
+    dump.create_named_pipe_for_dump(dump_filename)
+    return execute_command(
+        dump_command,
+        extra_environment=extra_environment,
+        run_to_completion=False,
+    )
 
 
 def use_streaming(databases, config, log_prefix):
@@ -237,44 +235,25 @@ def restore_data_source_dump(
         shlex.quote(part) for part in shlex.split(data_source.get('mysql_command') or 'mysql')
     )
 
-    dump_format = data_source.get('format')
-    is_directory_format = dump_format == 'directory'
+    is_directory_format = data_source.get('format') == 'directory'
 
-    if is_directory_format:
-        mysql_restore_command = tuple(
-            shlex.quote(part) for part in shlex.split(data_source.get('mysql_command') or 'mysql')
+    restore_command = (
+        mysql_restore_command
+        + (
+            tuple(data_source['restore_options'].split(' '))
+            if 'restore_options' in data_source
+            else ()
         )
-        restore_command = (
-            mysql_restore_command
-            + ('--batch',)
-            + (
-                tuple(data_source['restore_options'].split(' '))
-                if 'restore_options' in data_source
-                else ()
-            )
-            + (('--host', hostname) if hostname else ())
-            + (('--port', str(port)) if port else ())
-            + (('--protocol', 'tcp') if hostname or port else ())
-            + (('--user', username) if username else ())
-        )
-    else:
-        restore_command = (
-            mysql_restore_command
-            + ('--batch',)
-            + (
-                tuple(data_source['restore_options'].split(' '))
-                if 'restore_options' in data_source
-                else ()
-            )
-            + (('--host', hostname) if hostname else ())
-            + (('--port', str(port)) if port else ())
-            + (('--protocol', 'tcp') if hostname or port else ())
-            + (('--user', username) if username else ())
-        )
+        + (('--host', hostname) if hostname else ())
+        + (('--port', str(port)) if port else ())
+        + (('--protocol', 'tcp') if hostname or port else ())
+        + (('--user', username) if username else ())
+    )
 
     extra_environment = {'MYSQL_PWD': password} if password else None
 
     logger.debug(f"{log_prefix}: Restoring MySQL database {data_source['name']}{dry_run_label}")
+
     if dry_run:
         return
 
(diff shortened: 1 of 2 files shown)


shivansh02:
To test the feature, I tried running `mysqldump` on a database but got the following error:
`The MySQL server is running with the --secure-file-priv option so it cannot execute this statement](https://stackoverflow.com/questions/74932524/mysql-error-code-1290-the-mysql-server-is-running-with-the-secure-file-priv)`

To fix that, I added the following lines to mysql config files:
```
[mysqld]
secure-file-priv = ""
```

Now, the error I'm getting is 
`unknown variable 'secure-file-priv='`
I removed the variable from all the config files I tried setting it in, the error still remains.
Tried to allow AppArmor according to this answer, but to no avail:
[https://stackoverflow.com/a/57502632/20831469](https://stackoverflow.com/a/57502632/20831469)

witten:
Maybe try `secure_file_priv` with underscores instead of dashes? Although [these docs](https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_secure_file_priv) say an empty value has "no effect" which is not reassuring. You could also try setting it to the actual directory that you'll be dumping to.

borgmatic/hooks/mysql.py
```diff
@@ -86,6 +86,10 @@ def execute_dump_command(
     mysql_dump_command = tuple(
         shlex.quote(part) for part in shlex.split(database.get('mysql_dump_command') or 'mysqldump')
     )
+
+    dump_format = database.get('format')
```
witten, line 90: Style nit: This variable could easily inlined given that it's not used anywhere else besides the next line.

borgmatic/hooks/mysql.py
```diff
@@ -105,13 +109,20 @@ def execute_dump_command(
...
-        run_to_completion=False,
-    )
+    if is_directory_format:
+        dump.create_parent_directory_for_dump(dump_filename)
+        execute_command(
+            dump_command,
+            extra_environment=extra_environment,
+        )
+        return None
+    else:
```
witten, line 119: You could lose the `else:` given that the other code path returns early. (Do not feel strongly.)

borgmatic/hooks/mysql.py
```diff
@@ -153,31 +164,31 @@ def dump_data_sources(databases, config, log_prefix, dry_run):
...
-        else:
-            processes.append(
-                execute_dump_command(
-                    database,
+                result = execute_dump_command(
+                    renamed_database,
                     log_prefix,
                     dump_path,
-                    dump_database_names,
+                    (dump_name,),
```
witten, line 171: I'm not quite following this changed and the earlier deletions. Don't you still need some of that logic if the database name is `all`?

borgmatic/hooks/mysql.py
```diff
@@ -225,31 +236,69 @@ def restore_data_source_dump(
...
     )
-    restore_command = (
-        mysql_restore_command
-        + ('--batch',)
-        + (
-            tuple(data_source['restore_options'].split(' '))
-            if 'restore_options' in data_source
-            else ()
+
+    dump_format = data_source.get('format')
```
witten, line 240: Similar here; could be inlined.

borgmatic/hooks/mysql.py
```diff
@@ -225,31 +236,69 @@ def restore_data_source_dump(
...
+            + (
+                tuple(data_source['restore_options'].split(' '))
+                if 'restore_options' in data_source
+                else ()
+            )
+            + (('--host', hostname) if hostname else ())
+            + (('--port', str(port)) if port else ())
+            + (('--protocol', 'tcp') if hostname or port else ())
+            + (('--user', username) if username else ())
+        )
```
witten, line 273: I think I'm missing something.. These two commands look identical to me. Maybe it's just still WIP.

witten (commented):
What do you think of making these changes to the MariaDB database hook as well, as it's pretty identical to the MySQL hook right now? The main difference is that the specific binaries it runs have different names.

borgmatic/hooks/mysql.py
```diff
@@ -225,31 +236,69 @@ def restore_data_source_dump(
...
-        extra_environment=extra_environment,
-    )
+    if is_directory_format:
+        dump_path = make_dump_path(config)
+        dump_directory = dump.make_data_source_dump_filename(
+            dump_path, data_source['name'], data_source.get('hostname')
+        )
+        for file in os.listdir(dump_directory):
+            file_path = os.path.join(dump_directory, file)
+            if file.endswith('.sql'):
```
witten, line 288: [The docs say]( https://dev.mysql.com/doc/refman/8.4/en/mysqldump.html#option_mysqldump_tab)these are `.txt` files...?

shivansh02:
> What do you think of making these changes to the MariaDB database hook as well, as it's pretty identical to the MySQL hook right now? The main difference is that the specific binaries it runs have different names.

Yes I proposed making similar changes to MariaDB hook as well. I'll get on it as soon as I can get this PR approved, so it saves time.

shivansh02:
I was able to fix the `--secure-file-priv` error, but when I run 
`sudo borgmatic --add-drop-database --databases DBName --result-file /root/.borgmatic/mysql_databases/localhost/DBName`,
I get the following error:
`Cant create/write to file /root/.borgmatic/mysql_databases/localhost/DBName (OS errno 20 - Not a directory).`
Am I missing something? I ran this command to debug why my DB dump was not present in my backups, even without the --tab option.

shivansh02:
@witten I made changes that hopefully resolve all the issues above. Please take a look.

borgmatic/hooks/mysql.py
```diff
@@ -238,18 +252,41 @@ def restore_data_source_dump(
...
+        if not os.path.exists(dump_directory):
+            logger.warning(f"{log_prefix}: Dump directory {dump_directory} does not exist.")
+            return
+
+        for filename in os.listdir(dump_directory):
+            if filename.endswith('.sql'):
+                file_path = os.path.join(dump_directory, filename)
+                logger.debug(f"{log_prefix}: Restoring from {file_path}")
+                with open(file_path, 'r') as sql_file:
+                    execute_command_with_processes(
```
witten, line 278: I don't think you want the `_with_processes()` variant for this use case. That's intended for passing in, say, an `extract_process` so as to extract a single file into a restore command via background streaming. But that's not what you're doing here; this is presumably a synchronous call (... it's not streaming from `borg extract`). So, for that, maybe try just `execute_command()`?

witten (changes requested):
Finally got around to giving this a look. Thank you so much for your patience! I'll try to be a little faster on the next round.

witten:
Do you still plan on working on this one? Thanks!

witten:
Closing due to inactivity, but I'd be happy to revisit if you're still game. Thanks!

PR #86, opened 2024-10-30
Added support for Pushover Hook for notifications

tony1661:
Hi,

This PR aims to add support for Pushover notifications.

I've updated the documentation, list of supported integrations on borgmatic website homepage and have included various tests.

Pushover uses two keys. A user key and an application key. User keys are how you are authenticated to your Pushover account. Application keys are how Pushover knows which application to post the message to. Applications can be though of as threads, or topics.

I've included support for all of the API options that Pushover has.

Please let me know if there are any changes I should make.


witten:
Thanks for taking the time to implement yet another monitoring hook! I'll have a look at this after the 1.9.0 release—which I'm currently trying to avoid scope creep on so that it's released sometime this century. :smile:

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+    logger.info(f'{config_filename}: Updating Pushover {dry_run_label}')
+
+    if token is None:
+        logger.warning(f'{config_filename}: Token missing for Pushover')
+        return
+    if user is None:
+        logger.warning(f'{config_filename}: User missing for Pushover')
+        return
+
+    if 'priority' in state_config and state_config['priority'] == 2:
```
witten, line 49: It would be good for magic numbers here and below to get factored out into global constants.

Do not feel super strongly though.

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+        if 'expire' not in state_config:
+            logger.info(f'{config_filename}: Setting expire to default (10min).')
+            state_config['expire'] = 1200
+        if 'retry' not in state_config:
+            logger.info(f'{config_filename}: Setting retry to default (30sec).')
+            state_config['retry'] = 30
+    else:
+        state_config.pop('expire', None)
+        state_config.pop('retry', None)
```
witten, line 58: So if I'm reading this section correctly, `expire` and `retry` are ignored unless `priority` is 2? And they're set to default values only when `priority` is 2? Rather than silently swallowing these values, it might be good to raise an error because it's basically a misconfiguration.

Another option would be to push this logic into the schema via [`oneOf`](https://json-schema.org/understanding-json-schema/reference/combining#oneOf) (e.g., so that there are two sub-schemas, one with `priority` with only the choice of 2 and another schema with `priority` only 1 or below). But that might be more complicated JSON Schema work than you want to undertake now.

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+
+    data = {
+        'token': token,
+        'user': user,
+        'message': state.name.lower(),  # default to state name. Can be overwritten in state_config loop below.
+    }
+    
+    for key in state_config:
+        data[key] = state_config[key]
```
witten, line 67: A more succinct way to accomplish this would be something like:

```python
data = dict(
    {
        'token': token,
        'user': user,
        'message': state.name.lower(),  # default to state name. Can be overwritten in state_config loop below.
    },
    **state_config,
)
```

Or maybe this is more your style:

```python
data = {
    'token': token,
    'user': user,
    'message': state.name.lower(),  # default to state name. Can be overwritten in state_config loop below.
}

data.update(state_config)

File: borgmatic/hooks/pushover.py [hunk 4]
```diff
@@ -0,0 +1,89 @@
...
+
+    if state.name.lower() not in run_states:
+        return
+
+    dry_run_label = ' (dry run; not actually updating)' if dry_run else ''
+
+    state_config = hook_config.get(
+        state.name.lower(),
+        {
+            'message': state.name.lower(),
```
witten, line 33: This seems unnecessary given that you also default `message` below...?

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,75 @@ If you have any issues with the integration, [please contact
 us](https://torsion.org/borgmatic/#support-and-contributing).
 
 
+## Pushover hook
+
+<span class="minilink minilink-addedin">New in version 1.9.0</span>
```
witten, line 296: This will be 1.9.1 or later, but I can always set that after merging.

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,75 @@ If you have any issues with the integration, [please contact
...
+[Pushover](https://pushover.net) makes it easy to get real-time notifications 
+on your Android, iPhone, iPad, and Desktop (Android Wear and Apple Watch, too!)
```
witten, line 298: Nit: Add period at the end of the sentence.

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,75 @@ If you have any issues with the integration, [please contact
...
+
+```yaml
+pushover:
+    token: 7ms6TXHpTokTou2P6x4SodDeentHRa
+    user: hwRwoWsXMBWwgrSecfa9EfPey55WSN
+```
+
+
+With this configuration, borgmatic creates a Pushover event for your service
+whenever backups fail, but only when any of the `create`, `prune`, `compact`,
```
witten, line 317: It might be a little clearer / more accurate to say "whenever borgmatic fails" since it's not only on backups.

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,75 @@ If you have any issues with the integration, [please contact
...
+or `check` actions are run. Note that borgmatic does not contact Pushover
+when a backup starts or when it ends without error.
```
witten, line 319: Might be good to add "by default" for this.

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,75 @@ If you have any issues with the integration, [please contact
...
+```yaml
+pushover:
+    token: 7ms6TXHpTokTou2P6x4SodDeentHRa
+    user: hwRwoWsXMBWwgrSecfa9EfPey55WSN
+    start:
+        message: "Backup <b>Started</b>"
+        priority: -2
+        title: "Backup Started"
+        html: 1
+        ttl: 10 # Message will be deleted after 10 seconds.
```
witten, line 336: Nit: Convention is two spaces before an inline comment's `#`.

witten (commented):
Initial review (not including tests). This looks like a great start! Most of my comments are on smaller stuff, so hopefully there shouldn't be too many changes. Thanks!

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+            user:
+                type: string
+                description: |
+                    Your user/group key (or that of your target user), viewable 
+                    when logged into your dashboard: often referred to as 
+                    USER_KEY in Pushover documentation and code examples.
+                example: hwRwoWsXMBWwgrSecfa9EfPey55WSN
+            start:
+                type: object
+                type: object
```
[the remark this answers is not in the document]
tony1661: fixed: https://github.com/borgmatic-collective/borgmatic/commit/a45ba8553c9f93ad1b6a5609cc66db8723e0533b

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                        example: A backup job has started.
+                    priority:
+                        type: integer
+                        description: |
+                            A value of -2, -1, 0 (default), 1 or 2 that indicates the message priority.
+                        example: 0
+                    expire:
+                        type: integer
+                        description: |
+                            he expire parameter specifies how many seconds
```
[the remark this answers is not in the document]
tony1661: agreed. Shorter is better: https://github.com/borgmatic-collective/borgmatic/commit/1177385e080e40bddd847937f49b3c7c22082e0d

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                        type: integer
+                        description: |
+                            The retry parameter specifies how often 
+                            (in seconds) the Pushover servers will send the 
+                            same notification to the user.
+                        example: 30
+                    device:
+                        type: string
+                        description: |
+                            The name of one of your devices to send just to that device instead of all devices.
```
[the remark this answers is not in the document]
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/009055c61a737900ac392e7e8b1d5f189d7d82f2

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+        },
+    )
+
+    token = hook_config.get('token')
+    user = hook_config.get('user')
+
+    logger.info(f'{config_filename}: Updating Pushover {dry_run_label}')
```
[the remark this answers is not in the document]
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/0fdee067c76ffce7c1fe186c806964a941123014

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                    USER_KEY in Pushover documentation and code examples.
+                example: hwRwoWsXMBWwgrSecfa9EfPey55WSN
+            start:
+                type: object
+                type: object
+                properties:
+                    message:
+                        type: string
+                        description: |
+                            Message to be sent to the user or group.
```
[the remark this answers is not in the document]
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/72b27b08582746e01d4cbb903d24273682998032

borgmatic/hooks/pushover.py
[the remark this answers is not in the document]
tony1661: agreed: https://github.com/borgmatic-collective/borgmatic/commit/290559116d2099c7e01dcdefddfa0c299f804717

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+        logger.warning(f'{config_filename}: Token missing for Pushover')
+        return
+    if user is None:
+        logger.warning(f'{config_filename}: User missing for Pushover')
+        return
+
+    if 'priority' in state_config and state_config['priority'] == 2:
+        if 'expire' not in state_config:
+            logger.info(f'{config_filename}: Setting expire to default (10min).')
+            state_config['expire'] = 1200
```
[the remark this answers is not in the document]
tony1661: you are right: https://github.com/borgmatic-collective/borgmatic/commit/85ea8f4f457931b4534a59048278ffbdce953cf2

borgmatic/hooks/pushover.py
tony1661: Hi, since expire and retry are only used on priority 2 (emergency), I wrote the logic to remove them if they are delared when the priority is not set to 2.

Would you be ok with me changing it to continue doing what it is doing however give a warning that says something like: "Expire and/or Retry should only be configured when setting Priority to 2. They will be removed from the request." ?

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,89 @@
...
+    user = hook_config.get('user')
+
+    logger.info(f'{config_filename}: Updating Pushover {dry_run_label}')
+
+    if token is None:
+        logger.warning(f'{config_filename}: Token missing for Pushover')
+        return
+    if user is None:
+        logger.warning(f'{config_filename}: User missing for Pushover')
+        return
```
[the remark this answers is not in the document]
witten: IMO yes, you can safely remove the tests. I assume they're not testing the requiredness in the schema anyway. But even if they were, the schema isn't code that needs testing IMO; it's just declarative.

Also, side note: I don't think you _need_ quotes in YAML like in your changeset (when it's just a single-word string).

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                            he expire parameter specifies how many seconds 
+                            your notification will continue to be retried 
+                            for (every retry seconds).
+                        example: 1200
+                    retry:
+                        type: integer
+                        description: |
+                            The retry parameter specifies how often 
+                            (in seconds) the Pushover servers will send the 
+                            same notification to the user.
```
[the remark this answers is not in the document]
witten: No, but it just looks like one of the lines is wrapped too early. For instance, when I put it in vim and wrap it at 80 chars:

>                            The retry parameter specifies how often (in seconds)
>                            the Pushover servers will send the same notification
>                            to the user.

So this wouldn't make tests error; it's just a cosmetic thing.

borgmatic/hooks/pushover.py
witten: That's fine with me, although what's the use case for allowing a user to specify `expire`/`retry` when `priority` != 2? Like why not just error?

tony1661: My thought was to make it a bit more forgiving to someone who is setting up the hook. Instead of not working, we'd warn them.

I don't feel strongly either way so I can change it if you'd like 😁

witten: I don't feel strongly but I generally err towards erroring on misconfigurations (as opposed to runtime issues), but I could see the merits of both approaches!

tony1661: I will keep with the standard practice then and error out. Logging an error and returning is ok?

witten: That actually won't cause execution to stop. So if you want to error instead of warn, I would instead suggest raising `ValueError`, which borgmatic will catch higher up and display nicely. See `normalize.py` if you want an example of raising that exception type for a similar purpose.

docs/how-to/monitor-your-backups.md
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/a4fabb85217e6e57007f32248a5e06eace62f048

docs/how-to/monitor-your-backups.md
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/7eb19cb0a7747b5062b9ed3f8ca1f9e67a4efe20

docs/how-to/monitor-your-backups.md
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/3b79482b242ecb76c9a638461b4be65b66bd35ce

docs/how-to/monitor-your-backups.md
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/4ba42e89053c52ac84489d849ed040ad254b3df8

docs/how-to/monitor-your-backups.md
tony1661: https://github.com/borgmatic-collective/borgmatic/commit/59f8722e05c30d1f4ef343237bfdb5f0101d6551

borgmatic/hooks/pushover.py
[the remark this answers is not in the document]
tony1661: removed tests: https://github.com/borgmatic-collective/borgmatic/commit/3e0fa578603ddd94f02ab2c6a24931eaca17cf68

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
tony1661: vscode shows me at 81 link that:
image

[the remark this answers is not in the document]
witten: Okay, well then maybe I've got my wrapping incorrect here. I guess don't worry about it for now!

borgmatic/hooks/pushover.py
tony1661: agreed: https://github.com/borgmatic-collective/borgmatic/commit/46ebb0cebbebc691d734e3f850a2ee1b7d480c68

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                            A value of -2, -1, 0 (default), 1 or 2 that indicates the message priority.
+                        example: 0
+                    expire:
+                        type: integer
+                        description: |
+                            he expire parameter specifies how many seconds 
+                            your notification will continue to be retried 
+                            for (every retry seconds).
+                        example: 1200
+                    retry:
```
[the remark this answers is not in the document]
tony1661: Let me know if that covers it 🙂

borgmatic/hooks/pushover.py
tony1661: I like that idea. Makes it easier to read: https://github.com/borgmatic-collective/borgmatic/commit/51b885e7db7001de0a63cefb604d7f15058c22d6

borgmatic/hooks/pushover.py
tony1661: I've added to the above commit to fix it breaking a few tests: https://github.com/borgmatic-collective/borgmatic/commit/ef66349674c10fec1b84ead129a4028c98340bb6

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: The defaults look good, but I was also talking about indicating in the description that `expire` and `retry` only apply when `priority` is 2.

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,430 @@
...
+        hook_config,
+        {},
+        'config.yaml',
+        borgmatic.hooks.monitor.State.START,
+        monitoring_log_level=1,
+        dry_run=False,
+    )
+
+
+def test_ping_monitor_start_state_backup_default_message_with_priority_emergency_declared_no_expiry_or_retry_success():
```
witten, line 111: Maybe this could be `..._with_priority_emergency_uses_expire_and_retry_defaults()` or something? Like it's implied they're not declared IMO.

Do not feel strongly.

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,430 @@
...
+
+
+def test_ping_monitor_start_state_backup_default_message_with_priority_emergency_declared_no_expire_with_retry_success():
+    # This simulates priority level 2  and retry being set but expire is
+    # not declared. This should set expire to the default.
+    hook_config = {
+        'token': 'ksdjfwoweijfvwoeifvjmwghagy92',
+        'user': '983hfe0of902lkjfa2amanfgui',
+        'states': {'start', 'fail', 'finish'},
+        'start': {'priority': 2, 'expire': 30},
```
witten, line 184: Shouldn't this one be `retry` instead of `expire`?

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,430 @@
...
+        hook_config,
+        {},
+        'config.yaml',
+        borgmatic.hooks.monitor.State.START,
+        monitoring_log_level=1,
+        dry_run=False,
+    )
+
+
+def test_ping_monitor_start_state_backup_default_message_with_priority_high_declared_expire_and_retry_delared_success():
```
witten, line 210: This one too: IMO it's not just the success you care about but the fact that expire and retry are ignored.

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,80 @@
...
+def ping_monitor(hook_config, config, config_filename, state, monitoring_log_level, dry_run):
+    '''
+    Post a message to the configured Pushover application.
+    If this is a dry run, then don't actually update anything.
+    '''
+
+    run_states = hook_config.get('states', ['fail'])
+
+    if state.name.lower() not in run_states:
+        return
```
witten, line 29: Super minor, but: Does this case have any test coverage? You can tell by running `tox` and looking at the coverage output near the bottom.. It'll tell you about any lines that have zero test coverage.

borgmatic/hooks/pushover.py
```diff
@@ -0,0 +1,80 @@
...
+    data = {
+        'token': token,
+        'user': user,
+        'message': state.name.lower(),  # default to state name. Can be overwritten in state_config loop below.
+    }
+
+    for key in state_config:
+        data[key] = state_config[key]
+
+    if not dry_run:
```
witten, line 60: A dry run test case would be nice.

witten (commented):
Tests look great! Thank you so much for doing them (and the docs too!). I bet the tests were easier to write this time around.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
tony1661: Let me know if this covers it: https://github.com/tony1661/borgmatic/commit/eb8f7e03291718f6ecb4d8421c3cf5c7192324d1

borgmatic/hooks/pushover.py
tony1661: that is very cool. I didn't know that was possible: https://github.com/tony1661/borgmatic/commit/b1af304125ff07bf65ac7eaa71f59e2cfff8a25f

Tests all pass with it so I will resolve

docs/how-to/monitor-your-backups.md
```diff
@@ -290,6 +291,76 @@ If you have any issues with the integration, [please contact
...
+
+```yaml
+pushover:
+    token: 7ms6TXHpTokTou2P6x4SodDeentHRa
+    user: hwRwoWsXMBWwgrSecfa9EfPey55WSN
+```
+
+
+With this configuration, borgmatic creates a Pushover event for your service
+whenever borgmatic fail, but only when any of the `create`, `prune`, `compact`,
```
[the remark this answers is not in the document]
tony1661: https://github.com/tony1661/borgmatic/commit/0441e79b41fe202151004c24d8cf9c8955589e0f

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,430 @@
...
+        monitoring_log_level=1,
+        dry_run=False,
+    )
+
+
+def test_ping_monitor_start_state_backup_default_message_successfully_send_to_pushover():
+    # This test should send a notification to Pushover on backup start
+    # since the state has been configured. It should default to sending
+    # the name of the state as the 'message' since it is not
+    # explicitly declared in the state config.
```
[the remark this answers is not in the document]
tony1661: https://github.com/tony1661/borgmatic/commit/6d9340ebb24de4c9e70ce1576447150ff3c18e5a

tests/unit/hooks/test_pushover.py
tony1661: fixed here: https://github.com/tony1661/borgmatic/commit/97949266b372c85e6953a64b3465769a5da98250 and here: https://github.com/tony1661/borgmatic/commit/96bb402837cd97e574a1ad6cd4a811a436976cd7

borgmatic/hooks/pushover.py
tony1661: https://github.com/tony1661/borgmatic/commit/4e1f256b483affde6289177b60f57213efe4f254

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: LGTM!

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,238 @@ properties:
...
+                            (in seconds) the Pushover servers will send the 
+                            same notification to the user.
+                        example: 30
+                    device:
+                        type: string
+                        description: |
+                            The name of one of your devices to send just to that device instead of all devices.
+                        example: pixel8
+                    html:
+                        type: integer
```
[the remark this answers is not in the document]
tony1661: I can do this however it means that I need to loop through my state_config to convert the `True` to a 1. Is that ok?

```python
state_config = {k: (1 if v is True else v) for k, v in state_config.items()}
```

borgmatic/hooks/pushover.py
tony1661: I didn't know that tox will show that. Very cool. I will figure out which missing test cases I need to add.

In the meantime, here is the test case requested: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/3ab4b45041a26cf09ff7b991cb3650bd50a6a43f

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: That sounds like it would work. (Although please expand [single-letter variables](https://torsion.org/borgmatic/docs/how-to/develop-on-borgmatic/#code-style).) But here's a Python trick that would make this a little easier!

```python
state_config = {key: int(value) for key, value in state_config.items()}
```

But, I hear you saying, that'll break on every field *other* than the boolean `html`! Okay, no problem:

```python
state_config = {key: int(value) for key if key == `html` else value, value in state_config.items()}
```

And then if you wanted it more declarative, you could even do it without modifying `state_config`. Example:

```python
    data = dict(
        {
            'token': token,
            'user': user,
            'message': state.name.lower(),  # default to state name. Can be overwritten in state_config loop below.
        },
        **{key: int(value) for key if key == `html` else value, value in state_config.items()},
    )
```

Do not feel strongly. Just giving you some options.

borgmatic/hooks/pushover.py
witten: Yeah it's the [`coverage`](https://coverage.readthedocs.io/) package that supports that. Pretty handy.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
tony1661: Done. Just needed some slight modifications but it's passing all tests: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/61e4eeff6ca9b7b0659960a306ba87879832b62a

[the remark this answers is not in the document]
witten: Thanks, but if I'm reading it correctly, I don't think this it handles the case where the `html` option is False. Doesn't that need to get converted to a `0`? That's why I was suggesting the `int()` trick.. It converts `True` to `1` and `False` to `0` so you don't have to implement that logic yourself!

borgmatic/hooks/pushover.py
tony1661: I found the missing tests. See here: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/ea47704d86b9c3ebe3e412144f1c0d4459e9b79b

```
---------- coverage: platform linux, python 3.12.3-final-0 -----------
Name    Stmts   Miss  Cover   Missing
-------------------------------------
TOTAL    3656      0   100%
```

borgmatic/hooks/pushover.py
tony1661: I am getting close. I have committed what seems to be working code https://github.com/borgmatic-collective/borgmatic/pull/86/commits/a939a66fb44421488332a5060164cefc7335fb50 ... however my test `test_ping_monitor_start_state_backup_default_message_with_priority_high_declared_expire_and_retry_delared_success` now fails. I cannot figure out how to get flexmock to look for a ValueError.

Any suggestions?

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
tony1661: Ah, I think I finally got it: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/5321301708186ea6733295e1279a8a68b105b0c5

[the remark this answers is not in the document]
tony1661: This should do it: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/5f3d4f9b03b6832781ebf071952a37a9afecfb97

borgmatic/hooks/pushover.py
witten: The magical incantation is:

```python
with pytest.raises(ValueError):
    module.ping_monitor(
        ...
    )
```

See `tests/unit/hooks/test_mariadb.py:test_dump_data_sources_errors_for_missing_all_databases()` for an example.

borgmatic/config/schema.yaml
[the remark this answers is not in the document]
witten: Looks good!

borgmatic/hooks/pushover.py
tony1661: Sorry for the delay. Here is the test. Thanks for the pointer, I was stuck on that since I wasn't sure pytest was allowed to be included and I kept trying to figure out a way with Flexmock to accomplish the task.

https://github.com/borgmatic-collective/borgmatic/pull/86/commits/dab0dfcb32a6658f0d57309b0a87edc62bd825da

tests/unit/hooks/test_pushover.py
tony1661: agreed. Hopefully this will do it: https://github.com/borgmatic-collective/borgmatic/pull/86/commits/25572c98d7337386bd875337608cc19ff2ccd278

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,537 @@
...
+        hook_config,
+        {},
+        'config.yaml',
+        borgmatic.hooks.monitor.State.START,
+        monitoring_log_level=1,
+        dry_run=False,
+    )
+
+
+def test_ping_monitor_start_state_backup_default_message_with_priority_high_declared_expire_and_retry_ignored_success():
```
witten, line 225: The "ignored" part of this is now out of date since this has been changed to raising.

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,537 @@
...
+    hook_config = {'token': 'ksdjfwoweijfvwoeifvjmwghagy92', 'user': '983hfe0of902lkjfa2amanfgui'}
+    flexmock(module.logger).should_receive('warning').never()
+    flexmock(module.requests).should_receive('post').with_args(
+        'https://api.pushover.net/1/messages.json',
+        headers={'Content-type': 'application/x-www-form-urlencoded'},
+        data={
+            'token': 'ksdjfwoweijfvwoeifvjmwghagy92',
+            'user': '983hfe0of902lkjfa2amanfgui',
+            'message': 'fail',
+        },
```
witten, line 462: I don't think you need `.with_args(...)` here, because you basically want an expectation that `post` is never called—regardless of the arguments it receives.

tests/unit/hooks/test_pushover.py
```diff
@@ -0,0 +1,537 @@
...
+    }
+    flexmock(module.logger).should_receive('warning').never()
+    flexmock(module.requests).should_receive('post').with_args(
+        'https://api.pushover.net/1/messages.json',
+        headers={'Content-type': 'application/x-www-form-urlencoded'},
+        data={
+            'token': 'ksdjfwoweijfvwoeifvjmwghagy92',
+            'user': '983hfe0of902lkjfa2amanfgui',
+            'message': 'start',
+        },
```
witten, line 491: Same here.

witten (approved):
This looks great! Thank you again for sticking with the PR. I'll merge this now and then fix the few remaining nits (noted in comments) so as to save another round of back-and-forth.

borgmatic/config/schema.yaml
```diff
@@ -1609,6 +1609,265 @@ properties:
...
+                        example: https://pushover.net/apps/xxxxx-borgbackup
+                    url_title:
+                        type: string
+                        description: |
+                            A title for the URL specified as the url parameter,
+                            otherwise just the URL is shown.
+                        example: Pushover Link
+            finish:
+                type: object
+                type: object
```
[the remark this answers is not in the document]
witten: FYI the [end-to-end tests](https://torsion.org/borgmatic/docs/how-to/develop-on-borgmatic/#end-to-end-tests) caught this even though the unit tests didn't! Might be something to keep in mind as a final check for next time, even if the e2e tests are more of a pain to run.

witten:
Released in borgmatic 1.9.2!

PR #89, opened 2025-02-13
Fix path handling error when handling btrfs '/' submodule

dmitry-t7ko:
Problem:

`borgmatic` is trying to remove directory '' (empty string), if configured to use btrfs snapshots, and data from '/' subvolume is included in the backup.

It would look like this with '--verbosity 2':

```
backupserver: Calling bootstrap hook function remove_data_source_dumps
backupserver: Looking for bootstrap manifest files to remove in /tmp/borgmatic-*/borgmatic/bootstrap
backupserver: Calling btrfs hook function remove_data_source_dumps
backupserver: findmnt -t btrfs --json --list
backupserver: Looking for snapshots to remove in /var/log/.borgmatic-snapshot-*/var/log
backupserver: Looking for snapshots to remove in /var/cache/.borgmatic-snapshot-*/var/cache
backupserver: Looking for snapshots to remove in /swap/.borgmatic-snapshot-*/swap
backupserver: Looking for snapshots to remove in /home/.borgmatic-snapshot-*/home
backupserver: Looking for snapshots to remove in /.borgmatic-snapshot-*
backupserver: Deleting Btrfs snapshot /.borgmatic-snapshot-108534
backupserver: btrfs subvolume delete /.borgmatic-snapshot-108534
Delete subvolume 329 (no-commit): '//.borgmatic-snapshot-108534'
backupserver: Error running actions for repository
backupserver: [Errno 2] No such file or directory: ''
```

occurs when:
- `btrfs:` section is present in config.
- data from `/` (like `/etc`) is backed up, or if bootstrapping is enabled.

Version: 
extra/borgmatic 1.9.10-1 (Arch)

### DIFF
--- a/borgmatic/hooks/data_source/btrfs.py
+++ b/borgmatic/hooks/data_source/btrfs.py
@@ -303,7 +303,6 @@ def remove_data_source_dumps(hook_config, config, borgmatic_runtime_directory, d
             snapshot_parent_dir = snapshot_path.rsplit(subvolume.path, 1)[0]
             if os.path.isdir(snapshot_parent_dir):
                 shutil.rmtree(snapshot_parent_dir)
-                continue
 
 
 def make_data_source_dump_patterns(
--- a/borgmatic/hooks/data_source/btrfs.py
+++ b/borgmatic/hooks/data_source/btrfs.py
@@ -299,9 +299,11 @@ def remove_data_source_dumps(hook_config, config, borgmatic_runtime_directory, d
                 logger.debug(error)
                 return
 
-            # Strip off the subvolume path from the end of the snapshot path and then delete the
-            # resulting directory.
-            shutil.rmtree(snapshot_path.rsplit(subvolume.path, 1)[0])
+            # Remove snapshot parent directory if it still exists (might not exist if snapshot was for '/')
+            snapshot_parent_dir = snapshot_path.rsplit(subvolume.path, 1)[0]
+            if os.path.isdir(snapshot_parent_dir):
+                shutil.rmtree(snapshot_parent_dir)
+                continue
 
 
 def make_data_source_dump_patterns(
--- a/tests/unit/hooks/data_source/test_btrfs.py
+++ b/tests/unit/hooks/data_source/test_btrfs.py
@@ -520,6 +520,18 @@ def test_remove_data_source_dumps_deletes_snapshots():
     flexmock(module).should_receive('delete_snapshot').with_args(
         'btrfs', '/mnt/subvol2/.borgmatic-5678/mnt/subvol2'
     ).never()
+    flexmock(module.os.path).should_receive('isdir').with_args(
+        '/mnt/subvol1/.borgmatic-1234'
+    ).and_return(True)
+    flexmock(module.os.path).should_receive('isdir').with_args(
+        '/mnt/subvol1/.borgmatic-5678'
+    ).and_return(True)
+    flexmock(module.os.path).should_receive('isdir').with_args(
+        '/mnt/subvol2/.borgmatic-1234'
+    ).and_return(True)
+    flexmock(module.os.path).should_receive('isdir').with_args(
+        '/mnt/subvol2/.borgmatic-5678'
+    ).and_return(True)
     flexmock(module.shutil).should_receive('rmtree').with_args(
         '/mnt/subvol1/.borgmatic-1234'
     ).once()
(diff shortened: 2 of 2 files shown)


witten (commented):
Thanks for taking the time to make these changes, write tests, *and* file this in the face of registration issues!

While this does look like an improvement over trying to remove a path that's an empty string (!!!), I'm not sure these changes go quite far enough. That's because, if I'm understanding the PR correctly, it's running `os.path.isdir()` on an empty string, which will always fail—and then it's not actually removing the `/.borgmatic-snapshot-1234` directory that's still hanging around.

So if we were actually to remove that directory, maybe something as dumb as this would work instead?

```
shutil.rmtree(snapshot_path.rsplit(subvolume.path, 1)[0] or snapshot_path)
```

Explanation:

 * For the "usual" non-root case like a snapshot path of `/mnt/subvolume/borgmatic-snapshot-1234/mnt/subvolume`, it'll correctly split off the final `/mnt/subvolume`, resulting in `/mnt/subvolume/borgmatic-snapshot-1234` getting removed from the filesystem.
 * For the `/` case like a snapshot path of `/.borgmaic-snapshot-1234`, the `.rsplit()` part will result in an empty string, which will fall back to actually removing the snapshot path via `or snapshot_path`.

I'm open to other approaches too if you feel like this is too magical, but I do think it's important to actually clean up after ourselves and delete that directory!

dmitry-t7ko:
Behavior that I see on my machine is snapshot is it's own directory, and `btrfs` snapshot deletion also removes it's own directory. After running patched version on my setup, I see no `.borgmatic-XXXXXX` leftovers in my root, thus there's nothing for `shutil.rmtree` to remove.

```
[21:26:40] /  
> ls -a /
./  ../  bin@  boot/  dev/  efi/  etc/  home/  lib@  lib64@  mnt/  .netrwhist  opt/  proc/  root/  run/  sbin@  .snapshots/  srv/  swap/  sys/  tmp/  usr/  var/

[21:26:41] /  
> sudo btrfs subvolume snapshot -r / /.borgmatic-test-snapshot

Create readonly snapshot of '/' in '/.borgmatic-test-snapshot'

[21:27:07]  /  
> ls -a /
./   bin@   .borgmatic-test-snapshot/  efi/  home/  lib64@  .netrwhist  proc/  run/   .snapshots/  swap/  tmp/  var/
../  boot/  dev/                       etc/  lib@   mnt/    opt/        root/  sbin@  srv/         sys/   usr/

[21:27:11] /  
> sudo btrfs subvolume delete /.borgmatic-test-snapshot
Delete subvolume 361 (no-commit): '//.borgmatic-test-snapshot'

[21:27:27]  /  
> ls -a /
./  ../  bin@  boot/  dev/  efi/  etc/  home/  lib@  lib64@  mnt/  .netrwhist  opt/  proc/  root/  run/  sbin@  .snapshots/  srv/  swap/  sys/  tmp/  usr/  var/
```

dmitry-t7ko:
Also, I think a similar issue with edge cases around path string handling that commit 120a29ab4dc3122511e3eba9f3e6eb3f63803407 dealt with. Maybe, similar change can be done here, but being new to the codebase, I wanted to avoid accidentally doing `rm -rf / --no-preserve-root`.

witten:
> Behavior that I see on my machine is snapshot is it's own directory, and btrfs snapshot deletion also removes it's own directory. After running patched version on my setup, I see no .borgmatic-XXXXXX leftovers in my root, thus there's nothing for shutil.rmtree to remove.

Ah, you are right of course! The part that I was missing is that with every subvolume path *other* than `/`, there's another directory structure created underneath the `.borgmatic-snapshot-1234` directory, as in `/mnt/subvolume/borgmatic-snapshot-1234/mnt/subvolume`. But when snapshotting `/`, the snapshot *is* only `/.borgmatic-snapshot-1234` and therefore borgmatic removes it and there's no further parent directory to remove.

So your solution seems like it works. I'm not convinced the `isdir()` is necessary though (e.g., using `if snapshot_parent_dir:` might be enough), but `isdir()` here certainly doesn't hurt anything.

> Also, I think a similar issue with edge cases around path string handling that commit https://github.com/borgmatic-collective/borgmatic/commit/120a29ab4dc3122511e3eba9f3e6eb3f63803407 and https://github.com/borgmatic-collective/borgmatic/commit/2467518d4e56e81846c4ea9c3ccb7d8a3cd51e4e
included (I'm looking at test_make_borg_snapshot_pattern_includes_slashdot_hack_and_stripped_pattern_path test case).

I'd be happy to have a look at it, but could you say exactly what function or line you're thinking about? Are you referring to `make_borg_snapshot_pattern()`? If so, could you say more about what the problem there might be?

> Maybe, similar change can be done here, but being new to the codebase, I wanted to avoid accidentally doing rm -rf / --no-preserve-root.

Probably wise to avoid that. :smile:

dmitry-t7ko:
> Ah, you are right of course! The part that I was missing is that with every subvolume path other than /, there's another directory structure created underneath the .borgmatic-snapshot-1234 directory, as in /mnt/subvolume/borgmatic-snapshot-1234/mnt/subvolume

Exactly, I could've conveyed that clearer. As a result, root subvolume becomes a special case, because `make_snapshot_path` rstrips trailing slashes. 

> So your solution seems like it works. I'm not convinced the `isdir()` is necessary though (e.g., using `if snapshot_parent_dir:` might be enough), but `isdir()` here certainly doesn't hurt anything.

My thinking with `isdir()` was that it's more robust with regards to trailing slash :

```python
> >>> "/.borgmatic-1234".rsplit("/", 1)[0]
''
>>> "/.borgmatic-1234/".rsplit("/", 1)[0]
'/.borgmatic-1234'
```
With just an empty string check borgmatic would still try to delete nonexistent folder.

> I'd be happy to have a look at it, but could you say exactly what function or line you're thinking about? Are you referring to make_borg_snapshot_pattern()? If so, could you say more about what the problem there might be?

I've taken another look at the code, and I think you can disregard my original comment. When I was going through the git log, one particular the slashdot hack, and one of the test cases has jumped out at me at test_btrfs.py:131. 
```
('/', Pattern('/'), Pattern('/.borgmatic-snapshot-1234/./'))
```
I interpreted this as another case where root btrfs subvolume required special handling, but not I'm not sure that's correct.

witten:
> With just an empty string check borgmatic would still try to delete nonexistent folder.

Got it. Okay, I'll merge this PR now.. Thanks again for catching and fixing the issue!

> I've taken another look at the code, and I think you can disregard my original comment. When I was going through the git log, one particular the slashdot hack, and one of the test cases has jumped out at me at test_btrfs.py:131.
> 
> `('/', Pattern('/'), Pattern('/.borgmatic-snapshot-1234/./'))`
>
> I interpreted this as another case where root btrfs subvolume required special handling, but not I'm not sure that's correct.

Yeah, I think this is one situation where a `/` subvolume path + pattern *isn't* actually a special case!

